Issue #165, opened 2014-12-08
Identifier set, but id sent


Caligone:
Hi there,

I started to use _ng-admin_ few days later to quick scaffold a nice GUI admin with an API built with Hapi and Mongoose. But I encountered an issue with the _Entity.identifier()_ method.

I have set my identifier field as following :

``` javascript
var user = new Entity('user').identifier(new Field('_id'));
```

But here is the request from the edit form :

```
Request URL: http://localhost:3000/v1/user/5482e8405e6d4bbf1d8656f1
Request Payload: 
    email: "usertest@mail.com"
    id: "5482e8405e6d4bbf1d8656f1"
    language: "548342cd7d4ada54720d9cc0"
    password: "password"
    pseudo: "UserTest"
```

As you can see the identifier is correctly set on the URL, but the payload still have the default one as `id` instead of `_id`.

osiloke:
So what was the fix? Is there an additional option?

PR #170, opened 2014-12-15
[RFR] Specific endpoints

### LINKED ISSUE #130: Be able to change baseApiUrl for Entity
In some cases we need to use different base api url for different entities.
Live example ( Parse.com ):

api.parse.com/1/users
api.parse.com/1/classes/CustomClass

Do we have this possibility in ng-admin?

manuquentin:
Fixes #130, #165, #171
- [x] Allow to set URL for `view` & baseUrl for `entity`.
- [x] Allow to set query params `view` & `entity`.
- [x] Use new URLs for listView
- [x] Use new URLs  for show, create, edit & delete view
- [x] Add unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,11 @@ app.config(function (NgAdminConfigurationProvider, Application, Entity, Field, R
 
     var app = new Application('ng-admin backend demo') // application main title
         .transformParams(function(params) {
-            // Override query params here
+            // Sort by title by default
+            if (typeof params._sort === 'undefined') {
+                params._sort = 'title';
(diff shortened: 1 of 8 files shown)


osiloke:
It seems changes from this branch were removed from the master branch. The fix for custom resource identifiers does not exist. Is this correct.

fzaninotto:
Which fix? Can you point to the commit / line?

osiloke:
Sorry i assumed there was an issue within ng-admin but it was an issue with my api.

PR #207, opened 2014-12-28
[RFR] Don't transit pagination through datagrid

fzaninotto:
We can simplify the datagrid code (and binding) by moving the pagination out.

Also, using one-way databinding for entries to save on memory.

Finally, the pagination markup can be simplified to use native bootstrap components.

image

### DIFF
--- a/src/javascripts/ng-admin/Crud/list/maDatagrid.js
+++ b/src/javascripts/ng-admin/Crud/list/maDatagrid.js
@@ -12,7 +12,7 @@ define(function (require) {
             template: datagridView,
             scope: {
                 name: '@',
-                entries: '&',
+                entries: '=',
                 fields: '&',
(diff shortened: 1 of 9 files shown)


src/javascripts/ng-admin/Crud/list/maDatagrid.js
```diff
@@ -12,14 +12,10 @@ define(function (require) {
             template: datagridView,
             scope: {
                 name: '@',
-                entries: '=',
+                entries: '&',
```
manuquentin: `entries` can't be in a one-way binding.
When doing infinite pagination, we have to update `entities` array to display next entities in the list.

fzaninotto: Of course, you're right.

PR #214, opened 2015-01-07
[RFR] Rich example

fzaninotto:
- [x] Improve the complexity and test data of the bundled blog example
- [x] Introduce and document View.fields() to allow fields reusability


fzaninotto:
Weird, protractor tests don't run anymore on Travis... Any idea @manuquentin @jeromemacias?

fzaninotto:
Got it: the problem comes from minification. I test the constructor name, uglify breaks it.

PR #215, opened 2015-01-08
retrieve identifiers before getSingleApiCall call

endel:
Moved identifiers retrieval before calling `getSingleApiCall` on `getReferencedValues` method.

Screenshot demonstrating the problem:
captura de tela 2015-01-07 as 22 44 50

### DIFF
--- a/src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
+++ b/src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
@@ -141,14 +141,13 @@ define(function (require) {
         for (i in references) {
             reference = references[i];
             referencedView = reference.getReferencedView();
+            identifiers = reference.getIdentifierValues(rawValues);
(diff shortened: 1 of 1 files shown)


fzaninotto:
Not sure about this one. @manuquentin, can you review the patch?

fzaninotto:
I was right to be not sure... This broke the edit view in the production version (not in the development version, and I'm not sure why). Reverted and fixed the problem properly in #220.

manuquentin:
I don't see the same code from the [diff](https://github.com/marmelab/ng-admin/pull/215/files#diff-ce7353609be85e7f7c397fe44afb0cebR148) and the code [on master](https://github.com/marmelab/ng-admin/blob/1c7756b6a6cbc38b0ec818eb2b87e0c97c317d85/src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js#L145).

`identifiers` on master can't be undefined, it's weird that the diff doesn't shows the actual code on master.

Issue #129, opened 2014-11-19
Strip tags when displaying wysiwyg content in list


fzaninotto:
In a previous PR (#122), I fixed the problem of wysiwyg fields being displayed as HTML the wrong way: by displaying the unescaped HTML. This prevents truncation, and exposes the admin user to XSS attacks.

The right way to do it is to display a version of the wysiwyg stripped of HTML tags, equivalent to the result of PHP's `trip_tags()` function.

fzaninotto:
Fixed in #218

PR #218, opened 2015-01-08
[RFR] Strip tags in list wysiwyg fields

jpetitcolas:
Refs #129.

- [x] Strip tags when displaying WYSIWYG fields in List/Dashboard views
- [x] Unit test filter


jpetitcolas:
I'm wondering if we shouldn't add a line break (or at least a space) when we deal with block tags? Indeed, in the first record, `Title` is glued to the next paragraph. Any opinion?

image

We can't replace tags by a simple `space`, otherwise we would get too much space in some cases such as:

``` A short <strong>phrase</strong>.``` => ``` A short phrase .```

fzaninotto:
Valid concern, but let's not be too smart for now. I guess PHP's `strip_tags()` has the same shortcomings, we can live with it for the time being.

fzaninotto:
It's not necessary to commit built files. It makes merging harder, and we regularly rebuild on master.

jpetitcolas:
Added unit tests and switching to RFR. I reverted the changes brought to `build/*` files.

fzaninotto:
Needs rebase

jpetitcolas:
Rebased

fzaninotto:
Thanks!

PR #226, opened 2015-01-13
[RFR] Move example to example directory

fzaninotto:
...and cleanup the src and root of example files


examples/marvel.js
```diff
@@ -1,347 +0,0 @@
-(function() {
-    "use strict";
```
fzaninotto: Unless you feel like updating this one (heavily outdated), I prefer to delete it for now.

jeromemacias:
Tests now passed

PR #238, opened 2015-01-16
Update API-mapping.md

narigua:
Just a little update on doc for API that respond with a 206 partial or 200, check for JSON lengh.

### DIFF
--- a/doc/API-mapping.md
+++ b/doc/API-mapping.md
@@ -105,10 +105,14 @@ Add the following response interceptor:
 app.config(function(RestangularProvider) {
     RestangularProvider.addResponseInterceptor(function(data, operation, what, url, response) {
         if (operation == "getList") {
-            var contentRange = response.headers('Content-Range');
-            response.totalCount = contentRange.split('/')[1];
-        }
+            var contentRange = response.headers('content-range');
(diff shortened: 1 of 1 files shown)


endel:
I don't agree with this. The previous example was just enough to describe how to integrate with your back-end.

fzaninotto:
So that means your API doesn't include the Content-Range header when there is no pagination? I'm not sure that's standards compliant, in which case I agree with @endel.

narigua:
It's a Symfony2 project RESTFUL API, with Symfony best practice, it's not my only mind.

In this example, with a content-range, it's not useful add to header too much information when you have a little response. But it's a choice ;)

fzaninotto:
I don't see any `Content-Range` header in FOSRestBundle, which bundle are you using?

narigua:
http://jmsyst.com/bundles/JMSSerializerBundle

fzaninotto:
I don't see any `Content-Range` header there either...

narigua:
I'll give you this tomorrow, when i'm on the project.

fzaninotto:
Any updates on this one? As I don't see where the `Content-Range` header is set "the standard way" in your library, I'm tempted to close your PR.

PR #242, opened 2015-01-19
Update readme: dashboardView section

Xennis:
* The `limit` field of the `dashboardView` is used in the example, but not listed in the documentation.
* Also it is not mentioned, that you can use `sortField` and `sortDir` fields on a `dashboardView` (Both are not part of the general view settings and the `dashboardView` is only described as an "another special view", so it's not clear that this fields form the `listView` can be used for an `dashboardView` too).


fzaninotto:
Thanks!

PR #245, opened 2015-01-21
[RFR] Fix datagrid in show view + fix default filter param on datagrids for edit and show view

jeromemacias:


fzaninotto:
I don't know what it fixes, could you describe the pb, or better, add a test?

PR #246, opened 2015-01-21
Add option to define the route on a `isDetailLink` field

Xennis:
By default the `isDetailLink` field is linked to the edition view (or to the show view, if it is linked to a read only entity). I think there a cases, where you don't want to jump form the list view to the edition view of a specific item. That's why I added the following:

`detailLinkRoute(string)`
Define the route for a link in the list view, i.e. `isDetailLink` is true. The default is `edit`, hence the link points to the edition view. The other option is `show` to point to the show view.

Let me know, if I should change or improve something.

### DIFF
--- a/src/javascripts/ng-admin/Main/component/service/config/Field.js
+++ b/src/javascripts/ng-admin/Main/component/service/config/Field.js
@@ -22,7 +22,7 @@ define(function (require) {
         format: 'yyyy-MM-dd',
         template: defaultValueTemplate,
         isDetailLink: false,
-        detailLinkRoute: '',
+        detailLinkRoute: 'edit',
         list: true,
         dashboard: true,
         validation: {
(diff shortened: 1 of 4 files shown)


src/javascripts/ng-admin/Main/component/service/config/Field.js
```diff
@@ -22,6 +22,7 @@ define(function (require) {
         format: 'yyyy-MM-dd',
         template: defaultValueTemplate,
         isDetailLink: false,
+        detailLinkRoute: '',
```
fzaninotto: just set the default value to 'edit' here and you won't need the addition line 50.

fzaninotto:
Good addition. I hope the little fix I asked will retrigger the e2e tests correctly, because SauceLabs seems to have failed for no reason on your last commit.

Xennis:
I fixed it, but Travis CI still failed. Do you have an idea why? It's strange, because I only changed the readme file in the third commit, where it failed for the first time.

Xennis:
The first error, which I found in the details of Travis CI (the numbers in the beginning are the line numbers):
```
750 Selenium standalone server started at http://127.0.0.2:55489/wd/hub
751  [launcher] Error: UnknownError: chrome not reachable
```

fzaninotto:
I really don't understand why it fails on Travis. I'll merge it.

PR #259, opened 2015-01-26
Fix the sorting of ReferencedList fields in the showView.

Xennis:
Fix the sorting of ReferencedList fields in the showView. #254 (commit bc7745e54b6a1308d3b753fd6a65364bcbc1cc3f) wasn't enough to really fix it. Now it's possible to sort the ReferencedList table by a column.


fzaninotto:
Right!

fzaninotto:
And thanks!

PR #262, opened 2015-01-27
[RFR] Fix Reference identifiers

manuquentin:


fzaninotto:
tests still fail...

manuquentin:
`grunt-json-server` uses `json-server` 0.4.2 that requires a [`_sortDir`](https://github.com/typicode/json-server/blob/v0.4.1/src/routes.js#L36) param.
The `json-server` cli on the ng-admin demo server has the version 0.5.1 with a [`_order`](https://github.com/typicode/json-server/blob/master/src/create-routes.js#L42) params.

I tried to update `grunt-json-server` to 0.5 but there is too many thing to change and I'll break some features.

I will downgrade the `json-server` cli on the server to the 0.4.* to be able to run  dev, test & demo with the same parameters.

PR #264, opened 2015-01-29
[RFR] Fix map() being called twice on the listView

fzaninotto:


src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
```diff
@@ -64,7 +64,6 @@ define(function (require) {
 
                 entries = view.mapEntries(response.data);
```
fzaninotto: `mapEntries()` calls `mapEntry()`, which calls `getMappedValue()`

PR #270, opened 2015-01-31
[RFR] Allow listActions on ReferencedList

### LINKED ISSUE #269: Allow listActions on a ReferencedList field
I think, it would be nice to have the possibility to use the `listActions` option from the `listView` also on a `ReferencedList` field. Hence it would be possible to jump form the `ReferencedList` table to the edtion, show and/or delete view of the corresponding item.

Since the directive `<ma-datagrid>` is used  for the `ReferencedList` too, this shouldn't be a big change in the existing code. I wasn't able to do it, probably because of a lack of knowledge about Angular binding and compiling issues.

In [maColumn.js](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/column/maColumn.js) from line 20 on: Instead of parsing an empty array for `listActions`  you would parse then the new added variable `listActions` from the `ReferencedList` instance.

```javascript
                if (scope.type == 'ReferencedList') {
                    // special case to avoid recursion
                    element.append(
                        '<ma-datagrid name="{{ field.getReferencedView().name() }}"' +
                             'entries="field.entries"' +
                             'fields="field.getReferencedView().fields()"' +
                             'entity="field.getReferencedView().entity"' +
                             'listActions="[]">' +
                        '</ma-datagrid>'
                    );
                    $compile(element.contents())(scope);
                    return;
                }
```

fzaninotto:
Closes #269

### DIFF
--- a/src/javascripts/ng-admin/Crud/column/maColumn.js
+++ b/src/javascripts/ng-admin/Crud/column/maColumn.js
@@ -22,9 +22,9 @@ define(function (require) {
                     element.append(
                         '<ma-datagrid name="{{ field.getReferencedView().name() }}"' +
                              'entries="field.entries"' +
-                             'fields="field.getReferencedView().fields() | orderElement"' +
-                             'entity="field.getReferencedView().entity"' +
-                             'listActions="[]">' +
+                             'fields="::field.getReferencedView().fields() | orderElement"' +
+                             'list-actions="::field.listActions()"' +
+                             'entity="::field.getReferencedView().entity">' +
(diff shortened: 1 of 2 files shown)


fzaninotto:
Note that the next major version will probably refactor that, because it doesn't make sense to replicate the `listView` API on the `ReferencedList`. The right solution is to let the `ReferencedList` take a `listView` as parameter. 

But that's a deeper refactoring, and it doesn't justify delaying this simple change.

PR #274, opened 2015-02-01
Update documentation

Xennis:
See commit messages

### DIFF
--- a/doc/API-mapping.md
+++ b/doc/API-mapping.md
@@ -49,6 +49,18 @@ app.config(function(RestangularProvider) {
 });
 ```
 
+If your API requires that you post and put data for instance inside of an `item` field instead directly, use Restangular request interceptor:
+```js
+app.config(function(RestangularProvider) {
+    RestangularProvider.addFullRequestInterceptor(function(element, operation, what, url, headers, params, httpConfig) {
+        if(operation == 'post' || operation == 'put') {
+            element = { item: element };
+        }
+        return { element: element };
+    });
+}
+```
+
 ## Pagination
 
 ng-admin assumes that your API accepts `_page` and `_perPage` query parameters to paginate lists:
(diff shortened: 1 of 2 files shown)


doc/API-mapping.md
```diff
@@ -171,6 +171,33 @@ Now, the API call URLs will look like:
 
 http://your.api.domain/entityName?q=foo&tag=bar
 
+
+## Nested request and response
+
+If your API returns the results in a field of an object instead directly, you can use Restangular response interceptor. For instance if you request http://your.api.domain/entityName and the API returns the result array in an `items` field instead directly:
+
+```js
+app.config(function(RestangularProvider) {
```
fzaninotto: This is already covered in the 'Entry Format', line 40.

doc/API-mapping.md
```diff
@@ -171,6 +171,33 @@ Now, the API call URLs will look like:
...
+    RestangularProvider.addResponseInterceptor(function(data, operation, what, url, response) {
+        if (operation == "getList" && what == 'entityName') {
+            data = data.items;
+        }
+        return data;
+});
+```
+
+If your API requires that you post and put data for instance inside of an `item` field instead directly, use Restangular request interceptor:
```
fzaninotto: Good, but you shouldn't add a new section about it, "Entry Format" is there for it.

Xennis:
> This is already covered in the 'Entry Format', line 40.

In line 40 the method with the `RestangularProvider.addElementTransformer` is explained, but not the method with the `RestangularProvider.addResponseInterceptor`, when your data looks like that:

```
GET http://your.api.domain/entityName
{
  meta: { totalCount: 40, page: 2, ... },
  items: { HERE IS THE DATA ARRAY }
}
```

. The response interceptor is used in the "Total Number of Results" section too, so you can combine both easily.


> shouldn't add a new section about it, "Entry Format" is there for it.

Okay

fzaninotto:
When your data looks like the code extract you pasted, this is exactly the case to use  `addElementTransformer` instead of `addResponseInterceptor`. Besides, I don't want to replicate the entire Restangular documentation in ng-admin.

Xennis:
Okay, good to know.

Xennis:
Okay, good to know.

So may you just merge my first commit.

fzaninotto:
I'm interested in the second part of the Mapping addition, about ResponseInterceptor. Can you please update your PR accordingly?

Xennis:
Of course. But I am confused, you wrote
> this is exactly the case to use addElementTransformer instead of add**ResponseInterceptor**

and now
>  second part of the Mapping addition, about **ResponseInterceptor**

Was is just a typo and you meant, that you want the part about the RequestInterceptor?

fzaninotto:
Yes, it was a typo, sorry. We don't need the addition about response interceptor, but we need the addition about FullRequestInterceptor.

Xennis:
Okay, here we go.

fzaninotto:
Sorry, needs rebase

fzaninotto:
Thanks!

PR #276, opened 2015-02-02
[RFR] Dashboard panel titles now link to list view

fzaninotto:
image

I also took advantage to fix column ordering in the dashboard, and to fix pointer on links for the entire application.


lukemarsh:
Is there a way to disabled this?

fzaninotto:
You can still override the templates (see https://github.com/marmelab/ng-admin/blob/master/doc/Theming.md). But why would you want to do that?

PR #277, opened 2015-02-02
[RFR] Fix infinite pagination not keeping filters

fzaninotto:


src/javascripts/ng-admin/Crud/list/DatagridPagination.html
```diff
@@ -1,9 +1,9 @@
-<div ng-if="paginationCtrl.displayPagination">
+<div ng-if="!paginationCtrl.infinite">
```
fzaninotto: Now we display the number of results even if there is no need to paginate.

src/javascripts/ng-admin/Crud/list/maDatagridPagination.js
```diff
@@ -18,33 +17,11 @@ define(function (require) {
             },
             template: paginationView,
             controllerAs: 'paginationCtrl',
-            controller: DatagridPaginationController,
-            link: function (scope, element, attrs, controller) {
```
fzaninotto: As it didn't make sense to have code both in the `link` function and in the controller, I moved this part in `DatagridPaginationController`.

PR #278, opened 2015-02-02
[RFR] Strip restangular methods from entries

### LINKED ISSUE #273: bug with wysiwyg field named "head"
this is what happens when you edit an entity that have a field named "head":

schermata 2015-02-01 alle 17 21 05

instead of the content of that field you see that snippet of code. in show view you don't see anything but get lots of errors in the js console, like this one:

```
 TypeError: undefined is not a function
    at a.$inject (http://localhost:3000/public/vendor/ng-admin/ng-admin.min.js:4:5623)
    at http://localhost:3000/public/components/angular/angular.min.js:194:11
    at l.$get.xa.(anonymous function) [as value] (http://localhost:3000/public/components/angular/angular.min.js:59:175)
    at hb.functionCall (http://localhost:3000/public/components/angular/angular.min.js:198:426)
    at Object.<anonymous> (http://localhost:3000/public/components/angular/angular.min.js:108:188)
    at l.$get.l.$digest (http://localhost:3000/public/components/angular/angular.min.js:123:401)
    at l.$get.l.$apply (http://localhost:3000/public/components/angular/angular.min.js:126:471)
    at http://localhost:3000/public/components/angular/angular.min.js:139:187
    at e (http://localhost:3000/public/components/angular/angular.min.js:40:339)
    at http://localhost:3000/public/components/angular/angular.min.js:44:375
TypeError: Cannot read property 'replace' of undefined
    at a.$inject (http://localhost:3000/public/vendor/ng-admin/ng-admin.min.js:4:5622)
    at http://localhost:3000/public/components/angular/angular.min.js:194:11
    at l.$get.xa.(anonymous function) [as value] (http://localhost:3000/public/components/angular/angular.min.js:59:175)
    at hb.functionCall (http://localhost:3000/public/components/angular/angular.min.js:198:426)
    at Object.<anonymous> (http://localhost:3000/public/components/angular/angular.min.js:108:188)
    at l.$get.l.$digest (http://localhost:3000/public/components/angular/angular.min.js:123:401)
    at l.$get.l.$apply (http://localhost:3000/public/components/angular/angular.min.js:126:471)
    at http://localhost:3000/public/components/angular/angular.min.js:139:187
    at e (http://localhost:3000/public/components/angular/angular.min.js:40:339)
    at http://localhost:3000/public/components/angular/angular.min.js:44:375
```

fzaninotto:
Angular [adds its own method to every entity fetched](https://github.com/mgonto/restangular/blob/91be8642f21dcc275454d986b4cf8570366b84bd/src/restangular.js#L1062). As they are not namespaced, this has side effects.

I removed them, but I'm not sure we don't use them. @manuquentin, any insight?

Fixes #273

### DIFF
--- a/src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
+++ b/src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
@@ -24,10 +24,12 @@ define(function (require) {
      * @returns {promise} (list of fields (with their values if set) & the entity name, label & id-
      */
     RetrieveQueries.prototype.getOne = function (view, entityId) {
-        return this.Restangular
+        var Restangular = this.Restangular;
+        return Restangular
             .oneUrl(view.entity.name(), this.config.getRouteFor(view, entityId))
             .get()
             .then(function (response) {
+                response.data = Restangular.stripRestangular(response.data);
                 return view.mapEntry(response.data);
             });
     };
@@ -113,9 +115,14 @@ define(function (require) {
         }
 
         // Get grid data
-        return this.Restangular
+        var Restangular = this.Restangular;
+        return Restangular
             .allUrl(listView.entity.name(), this.config.getRouteFor(listView))
-            .getList(params);
+            .getList(params)
+            .then(function(response) {
+                response.data = Restangular.stripRestangular(response.data);
+                return response;
+            });
     };
 
     /**


src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
```diff
@@ -24,10 +24,12 @@ define(function (require) {
      * @returns {promise} (list of fields (with their values if set) & the entity name, label & id-
      */
     RetrieveQueries.prototype.getOne = function (view, entityId) {
-        return this.Restangular
+        var Restangular = this.Restangular;
+        return Restangular
             .oneUrl(view.entity.name(), this.config.getRouteFor(view, entityId))
             .get()
             .then(function (response) {
+                response.data = Restangular.stripRestangular(response.data);
```
manuquentin: We can maybe put this in the `view.mapEntry` method ?

manuquentin:
Restangular methods aren't used on the object himself, we can remove them

fzaninotto:
As a matter of fact, my patch doesn't solve #273, because properties named like Restangular methods (like "head") are added then removed, so there is no way to read them. We need a way to prevent Restangular from decorating them in the first place.

fzaninotto:
We can't prevent Restangular from decorating, but we can get the original entity by adding a listener (see https://github.com/mgonto/restangular/issues/100#issuecomment-19174352). Back to WIP.

fzaninotto:
I've fixed the bug, but there are two drawbacks:

- This breaks all tests using Restangular (long and cumbersome to fix)
- This slows down the app, as we have to clone all entries in order

So this needs more work.

PR #284, opened 2015-02-03
[RFR] Include bind polyfill

manuquentin:
[bind polyfill](https://github.com/marmelab/ng-admin/blob/fcb7666f1a93f0321ccfd2d86cdf0711082d0ce5/src/javascripts/test/karma.conf.js#L22) is included in test because travis runs PhantomJS that doesn't implement it.

ng-admin should be compatible with [browsers that don't support `bind`](http://caniuse.com/#search=bind), it should be included in the minified build.


fzaninotto:
[`function.prototype.build` is widely supported](http://kangax.github.io/compat-table/es5/#Function.prototype.bind). PhantomJS is the only one not supporting it, so I suggest you add the polyfill to your test build.

PR #286, opened 2015-02-03
adds option for patch updates

rall:
would something like this work for you? If this looks promising I'll update the tests to cover this additional case.

### DIFF
--- a/src/javascripts/ng-admin/Crud/repository/UpdateQueries.js
+++ b/src/javascripts/ng-admin/Crud/repository/UpdateQueries.js
@@ -25,7 +25,7 @@ define(function (require) {
      */
     UpdateQueries.prototype.updateOne = function (view, rawEntity) {
         var entityId = rawEntity[view.getEntity().identifier().name()],
-            method = view.getEntity().updateMethod(),
+            method = view.updateMethod(),
             url = this.Restangular.oneUrl(view.entity.name(), this.config.getRouteFor(view, entityId));
         var operation = method ? url.customOperation(method, null, {}, {}, rawEntity) : url.customPUT(rawEntity);
         return operation.then(function (response) {
--- a/src/javascripts/ng-admin/Crud/repository/UpdateQueries.js
+++ b/src/javascripts/ng-admin/Crud/repository/UpdateQueries.js
@@ -25,7 +25,7 @@ define(function (require) {
      */
     UpdateQueries.prototype.updateOne = function (view, rawEntity) {
         var entityId = rawEntity[view.getEntity().identifier().name()],
-            method = view.updateMethod(),
+            method = view.config.updateMethod,
             url = this.Restangular.oneUrl(view.entity.name(), this.config.getRouteFor(view, entityId));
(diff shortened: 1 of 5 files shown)


fzaninotto:
That's a good start, except that PATCH is supposed to send... a JSON patch, not an JSON entity. So the body format isn't the same at all.

rall:
ok, how about this? Moved updateMethod to the View config, allowing us to use different methods in the  create and edit views. It didn't really make sense to make that configuration in the Entity context now that I think about.

As for the payload, it seems the rawEntity argument comes from FormController.prototype.validate, and that returns whatever entity data is exposed in the view. If the view shows a subset of the entity, and it's configured to use patch then I think all is well.

rall:
any comments on this updated pull request?

src/javascripts/ng-admin/Crud/repository/UpdateQueries.js
```diff
@@ -24,15 +24,13 @@ define(function (require) {
...
-
-        // Get element data
-        return this.Restangular
-            .oneUrl(view.entity.name(), this.config.getRouteFor(view, entityId))
-            .customPUT(rawEntity)
-            .then(function (response) {
-                return view.mapEntry(response.data);
-            });
+        var entityId = rawEntity[view.getEntity().identifier().name()],
+            method = view.config.updateMethod,
```
fzaninotto: don't access `view.config` directly. Instead, please use `view.updateMethod()`

fzaninotto:
I don't understand why you moved `updateMethod` to the parent class, while it's only used in update scenario (in the Edition view). In fact, if you want to move it up, you should move it to the Entity, not the View.

The Entity class represents the mapping between the API and in-memory objects. View classes are related to the representation of these in-memory objects in the CRUD.

rall:
Thanks for the feedback. I'll try to explain my choices:

1. I moved the config to the parent because it's not specific to PATCH and the 'Edition' view - for example, it could equally apply to record creation, where one might want to pick POST or PUT. In fact I probably should have called the config "actionMethod" or something - less misleading that way

2. I moved the config _to_ view _from_ entity to answer your earlier comment: since a PATCH should represent a subset of the entire object, that subset can be defined by the fields of a view. 

Thoughts? If I revert back to updateMethod on the Entity config, what else would need to change?

rall:
Here's another proposal!

The update method config is back on Entity where it makes sense. In the updateOne method on UpdateQueries I'm checking which method we're going to use. If it's PATCH, then just send those properties of the entity that are displayed in the view, if it's anything else then send the whole entity.

Again, this obviously needs tests. I'll add them if you think I'm on the right path.

rall:
Thinking about this some more, it seems Restangular makes an OPTIONS request. I wonder how we might access the Access-Control-Allow-Methods header from that request? If it includes 'PATCH' then use that method for updates

fzaninotto:
I don't think your `pick` makes much sense here - if you should pick fields, it's not the ones displayed, it's the ones modified. You're making too many assumptions I guess. 

Your first commit was the closest to something mergeable I think.

fzaninotto:
No news for too long, closing.

PR #288, opened 2015-02-04
[RFR] Date format display

ThieryMichel:
Allow to configure separate date display format from stored one.


fzaninotto:
I think the 'format' function should pass from the model (a Date object) to the display, while a 'parse' function should pass from the display (a string) to the model (a date object, or a string formatted for the persistence).

ThieryMichel:
Comment applied

src/javascripts/ng-admin/Main/component/service/config/Field.js
```diff
@@ -19,6 +19,7 @@ define(function (require) {
         order: null,
         identifier: false,
         format: 'yyyy-MM-dd',
+        parse: null,
```
fzaninotto: This can't be null, or it will break everything. The default value for `parse` should be `function(x) { return x; }`.

src/javascripts/ng-admin/Crud/form/FormController.js
```diff
@@ -52,7 +52,7 @@ define(function () {
             field = fields[i];
             value = entry.values[field.name()];
             if (field.type() === 'date') {
-                value = $filter('date')(value, field.format());
+                value = $filter('date')(value, field.parse() || field.format());
```
fzaninotto: what if `field.parse()` returns `false` or 0 by design?

src/javascripts/ng-admin/Crud/filter/maFilterController.js
```diff
@@ -45,7 +45,7 @@ define(function () {
                 values[fieldName] = this.values[fieldName];
 
                 if (field.type() === 'date') {
-                    values[fieldName] = this.$filter('date')(values[fieldName], field.format());
+                    values[fieldName] = this.$filter('date')(values[fieldName], field.parse() || field.format());
```
fzaninotto: not good, see my comment below.

fzaninotto:
Needs rebase

src/javascripts/ng-admin/Main/component/service/config/Field.js
```diff
@@ -19,6 +19,9 @@ define(function (require) {
         order: null,
         identifier: false,
         format: 'yyyy-MM-dd',
+        parse: function (date) {
+            return date;
```
fzaninotto: Shouldn't it create a `Date` object by default? That way, when serialized, it would be sent as ISO.

fzaninotto:
Bump

src/javascripts/ng-admin/Main/component/service/config/Field.js
ThieryMichel: The date picker already return a date object.

ThieryMichel:
rebased

fzaninotto:
Thanks!

PR #293, opened 2015-02-05
Compile description field

Xennis:
Compile the `description` like the `title`. That means you can customize both with details from the current entry.

For instance you could display the ID of an entry in the `title` and the name in the `description`, because the names are too long for a big heading. Or you want to display some other extra information about the entry in the description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ The title of the view. ng-admin sees it as a template, and compiles it with the 
         editionView.title('Edit item "{{ entry.values.title }}"');
 
 * `description(String)`
-A text displayed below the title.
+A text displayed below the title. Like the `title` ng-admin sees it as a template and it can be customized in the same way.
 
 * `actions(String|Array)`
 Customize the list of actions for this view. You can pass a list of button names among 'back', 'list', 'show', create', 'edit', 'delete':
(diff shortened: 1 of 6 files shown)


fzaninotto:
tests fail

Xennis:
Do you know, which test failed? When I run `make test` it says "Executed 102 of 102 SUCESS" and "15 test, 34 assertions, 0 failures", so it seems to be okay. And Travis CI just outputs 

```
Running "protractor:e2e" (protractor) task
Starting selenium standalone server...
Selenium standalone server started at http://127.0.0.2:41818/wd/hub
[launcher] Error: UnknownError: chrome not reachable
```

fzaninotto:
Functional/e2e tests (the ones running scenarios in the browser) fail. I'm surprised they don't run when you call `make test`. What is the output of this command on your side?

Xennis:
Maybe I do something wrong on my commits? Because until now each of my commits failed on Travis CI with the strange error message `Error: UnknownError: chrome not reachable`. As I already wrote above the e2e tests output on my pc "15 test, 34 assertions, 0 failures":

capture

fzaninotto:
Tests pass on my side as well, Travis is drunk. Merging.

Thanks for the patch!

Issue #294, opened 2015-02-05
[RFC] Field Factories


fzaninotto:
# RFC: Field Factories

Now that's 0.5 is out, it's time to think about 1.0. There is a breaking change in the Configuration API that I think is necessary, and we should do that before 1.0 is released. Read on to understand the rationale.

## Problems

1. We add more and more methods to the generic `Field` class that are specific to a certain field type (e.g. `choices()`, `uploadInformation()`)
2. We can't add custom fields easily (forced to use the `template`) field
3. We can't override an existing field by a custom one
4. We can't use a different directive for the ShowView and for the ListView
5. We have three classes for references, but not yet any ability to define variants for view (autocomplete, options list, etc.)
6. We have to inject a lot of classes to the configuration function (`app.config(function (NgAdminConfigurationProvider, Application, Entity, Field, Reference, ReferencedList, ReferenceMany) { ... })`)

## Proposal

- Introduce a factory object to replace all calls to `new`

```js
// before
app.config(function (NgAdminConfigurationProvider, Application, Entity, Field) {
    var admin = new Application('ng-admin backend demo')
        .baseApiUrl('http://localhost:3000/');
    var post = new Entity('posts');
    post.listView()
        .fields([
            new Field('title'),
            new Field('is_published').type('boolean')
        ]);
    admin.addEntity(post);
    NgAdminConfigurationProvider.configure(admin);
});

// after
app.config(function (NgAdminConfigurationProvider) {
    var nga = NgAdminConfigurationProvider;
    var admin = nga.application('ng-admin backend demo')
        .baseApiUrl('http://localhost:3000/');
    var post = nga.entity('posts'); /
    post.listView()
        .fields([
            nga.field('title'),
            nga.field('is_published', 'boolean')
        ]);
    admin.addEntity(post);
    nga.configure(admin);
});
```

- `nga.field` instanciates subclasses of `Field` based on the type (which explains why the type must be in the signature).

```js
nga.field('title') // new TextField('title')
nga.field('is_published', 'boolean') // new BooleanField('is_published')
```

- Create subclasses of the `Field` class. Each subclass can add any field-specific method without cluttering the main `Field` class.

```js
// es6 syntax, could be es5 prototype, doesn't matter for now.
class NumberField extends TextField {
    format(formatString) {
        return ... // format number based on string
    }
}
```

- Store the directive to be used in each view in `FieldView` classes and subclasses.

```js
[...]

Issue #250, opened 2015-01-22
Lazy-load heavy dependencies


fzaninotto:
Currently, all the code necessary for all features is compiled into the `ng-admin.min.js` and `ng-admin.min.css` files. These files are already BIG (the JS is 350kB... and counting). As we add more sophisticated features (rich text editor, JSON editor), the code will grow... even if developers don't use these features. 

We need to be more conservative on the compiled file size, and let ng-admin load additional scripts when needed. A combination of something like [angular-load](https://github.com/urish/angular-load/blob/master/angular-load.js) and compiled files for "plugins" should do the trick.

MarZab:
Or letting the developer load the modules he needs, something like ngStrap aproach, load all if you dont care or load just the parts that you need.

Related to this, I see the CSS is compiled from Bootstrap-SASS, FontAwesome and SB-Admin-2. Would it be sensible to just include them from bower, that way one could use the existing (custom) boostrap and would lead the way to custom ng-admin templates :)

jpetitcolas:
Closing this issue as we refactored recently the whole build process, exposing two flavors of ng-admin:

* Standalone version, embedding all the dependencies in `ng-admin.min.js`
* Lite version including only `ng-admin` in `ng-admin-only.min.js`

PR #297, opened 2015-02-06
[WIP] Add datatimepicker field

manuquentin:
- [x] Add field
- [ ] Add doc

### DIFF
--- a/src/javascripts/ng-admin.js
+++ b/src/javascripts/ng-admin.js
@@ -11,9 +11,12 @@ require.config({
         'angular-resource': 'bower_components/angular-resource/angular-resource',
         'angular-sanitize': 'bower_components/angular-sanitize/angular-sanitize',
         'angular-ui-router': 'bower_components/angular-ui-router/release/angular-ui-router',
+        'bootstrap': 'bower_components/bootstrap/dist/js/bootstrap',
         'ng-file-upload': 'bower_components/ng-file-upload/angular-file-upload',
         'lodash': 'bower_components/lodash/dist/lodash.min',
         'text' : 'bower_components/requirejs-text/text',
+        'jquery' : 'bower_components/jquery/dist/jquery',
+        'moment' : 'bower_components/moment/min/moment.min',
         'angular-bootstrap': 'bower_components/angular-bootstrap/ui-bootstrap.min',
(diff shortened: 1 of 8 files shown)


fzaninotto:
So far, so good. I'm worried about the kB this is going to add to `ng-admin.min.js`, we really need to consider #250 soon.

src/javascripts/ng-admin.js
```diff
@@ -11,9 +11,12 @@ require.config({
         'angular-resource': 'bower_components/angular-resource/angular-resource',
         'angular-sanitize': 'bower_components/angular-sanitize/angular-sanitize',
         'angular-ui-router': 'bower_components/angular-ui-router/release/angular-ui-router',
+        'bootstrap': 'bower_components/bootstrap/dist/js/bootstrap',
```
fzaninotto: seems that only bootstrap transitions and collapse are needed at this point (http://eonasdan.github.io/bootstrap-datetimepicker/Installing/).

src/javascripts/ng-admin/Crud/field/maDateTimeField.js
```diff
@@ -0,0 +1,58 @@
...
+    'use strict';
+
+    var datetimepicker = require('bootstrap-datetimepicker');
+
+    /**
+     * Edition field for a date with time - a text input with a datepicker.
+     *
+     * @example <ma-date-time-field field="field" value="value"></ma-date-time-field>
+     */
+    function maDateField() {
```
fzaninotto: should probably be renamed to `maDateTimeField`

fzaninotto:
Seing the number of added dependencies (jQuery, moment, bootstrap js), I wonder if it's not better to add https://github.com/zhaber/datetimepicker (even though it has inferior UI). Your datetimepicker, better but heavier, could be provided as an extension... Once we know how to handle them (cf #294). 

What do you think?

manuquentin:
I've spent 3 hours trying to use [zhaber/datetimepicker](https://github.com/zhaber/datetimepicker) but apparently [it's not compatible with angular 1.3](https://github.com/zhaber/datetimepicker/issues/15).

I'll search for another directive.

fzaninotto:
Ah, bummer. Thanks for trying.

fzaninotto:
Any progress for documentation ?

fzaninotto:
Superseded by #348

PR #313, opened 2015-02-13
[RFR] allow actions and listActions to accept template in array

### LINKED ISSUE #302: listActions enhancement (mix custom and normal actions)
Actually we can pass a list of button names among 'show', 'edit', and 'delete' to listActions, or we can pass a template string to add custom actions. 
But if we want to mix both normal actions and custom one, we need to pass a template with the normal actions directive mixed with the custom one.
What I propose is to modify listAction, to accept both action name and template in its array form.
```
.listActions(['edit', '<my-custom-directive></my-custom-directive>'])
```

ThieryMichel:
fix #302


fzaninotto:
Is it refs #302?

ThieryMichel:
yes

PR #314, opened 2015-02-13
removing humane and replacing with alertify

lukemarsh:
Humane doesn't work well with require and causes unit tests to fail. I've replaced it with alertify.


fzaninotto:
I don't understand. Humane works fine on many projects using ng-admin, and tests pass (but not yours).

Did you `bower cache clean`?

fzaninotto:
No feedback for 9 days, closing.

PR #323, opened 2015-02-19
[RFR] Add possibility to filter references results

jeromemacias:
We can now apply filters on reference fields:
```js
nga.field('post_id', 'reference')
   .targetEntity(nga.entity('post'))
   .targetField(nga.field('title', 'string')))
   .filters({ 'is_published': true })
```

### DIFF
--- a/src/javascripts/ng-admin/Main/component/service/config/Reference.js
+++ b/src/javascripts/ng-admin/Main/component/service/config/Reference.js
@@ -15,7 +15,7 @@ define(function (require) {
         targetField : null,
         sortField: 'id',
         sortDir: 'DESC',
-        filters: {},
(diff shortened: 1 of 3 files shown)


src/javascripts/ng-admin/Main/component/service/config/Reference.js
```diff
@@ -15,6 +15,7 @@ define(function (require) {
         targetField : null,
         sortField: 'id',
         sortDir: 'DESC',
+        filters: {},
```
fzaninotto: Shouldn't you use `false`as default value to be iso with existing feature?

fzaninotto:
:+1:  Great idea!

PR #325, opened 2015-02-20
[RFR] Factory for field Views

fzaninotto:
Following #307, and refs #294, this PR introduces the second part of the factories: widget factories.

A widget factory looks like the following:

```js
define(function(require) {
    "use strict";

    function getReadWidget() {
        return '<ma-string-column value="::entry.values[field.name()]"></ma-string-column>';
    }
    function getLinkWidget() {
        return '<a ng-click="gotoDetail()">' + getReadWidget() + '</a>';
    }
    function getFilterWidget() {
        return '<ma-input-field field="::field" value="values[field.name()]"></ma-input-field>';
    }
    function getWriteWidget() {
        return '<ma-input-field type="email" field="::field" value="entry.values[field.name()]"></ma-input-field>';
    }
    return {
        getReadWidget:   getReadWidget,
        getLinkWidget:   getLinkWidget,
        getFilterWidget: getFilterWidget,
        getWriteWidget:  getWriteWidget,
    }
});
```

I didn't make a difference between listView and showView (refs #261), this will be in another PR.

Tasks:
- [x] Create configuration provider for widgets
- [x] Use widgets from config in listView
- [x] Use widgets from config in filters
- [x] Use widgets from config in creationView
- [x] Use widgets from config in editionView
- [x] Fix rebase
- [x] Fix e2e tests
- [x] Add doc

PR #327, opened 2015-02-24
[BC Break][RFR] Specialize Configuration Field Classes

fzaninotto:
This finishes the Factory refactoring by using specialized configuration field classes under the hood.

`new Field().type('date')`doesn't work anymore after this patch is merged, it breaks compatibility with 0.5.

Follows #325 and refs #294


fzaninotto:
Switching to RFR

Issue #322, opened 2015-02-19
Require.js problem on fresh installation


binarious:
## Problem

```
 Uncaught Error: Mismatched anonymous define() module: function (module) {  in require.js:166
````

## Steps to reproduce     

- New angular app: `yo angular`
  - 3x `yes` (Sass, Bootstrap, Bootstrap Sass)
  - include all modules
- Install ng-admin: `bower install ng-admin --save`
  - Select `1.3.*` at the version conflict
- Run the server: `grunt serve`
- Check the console

Any ideas what I did wrong?

fzaninotto:
This completely fails on my laptop with errors unrelated to ng-admin. I have no idea what's wrong with this installer...

fzaninotto:
So, I managed to reproduce the error. It seems that the generator you use includes all dependencies required by bower as third-party scripts:

```html
<!-- Present by default -->
    <script src="bower_components/jquery/dist/jquery.js"></script>
    <script src="bower_components/angular/angular.js"></script>
    <script src="bower_components/bootstrap-sass-official/assets/javascripts/bootstrap.js"></script>
    <script src="bower_components/angular-animate/angular-animate.js"></script>
    <script src="bower_components/angular-aria/angular-aria.js"></script>
    <script src="bower_components/angular-cookies/angular-cookies.js"></script>
    <script src="bower_components/angular-messages/angular-messages.js"></script>
    <script src="bower_components/angular-resource/angular-resource.js"></script>
    <script src="bower_components/angular-route/angular-route.js"></script>
    <script src="bower_components/angular-sanitize/angular-sanitize.js"></script>
    <script src="bower_components/angular-touch/angular-touch.js"></script>
<!-- Added by simply adding ng-admin using npm -->
    <script src="bower_components/jquery/dist/jquery.js"></script>
    <script src="bower_components/angular-bootstrap/ui-bootstrap-tpls.js"></script>
    <script src="bower_components/angular-ui-codemirror/ui-codemirror.js"></script>
    <script src="bower_components/angular-ui-router/release/angular-ui-router.js"></script>
    <script src="bower_components/bootstrap/dist/js/bootstrap.js"></script>
    <script src="bower_components/codemirror/lib/codemirror.js"></script>
    <script src="bower_components/es5-shim/es5-shim.js"></script>
    <script src="bower_components/humane/humane.min.js"></script>
    <script src="bower_components/inflection/inflection.min.js"></script>
[...]

PR #333, opened 2015-02-27
[RFR] Exclude dependencies during build

manuquentin:
- [x] Adding a build task to generate a `ng-admin-only.min.js` file without dependencies
- [x] Mocks requires depencies
- [x] Add documentation

Refs #322 

```sh
155K ng-admin-only.min.js
600K ng-admin.min.js
```


bower.json
```diff
@@ -6,34 +6,34 @@
     "build/ng-admin.min.js"
   ],
   "dependencies": {
-    "angular": "~1.3.1",
+    "angular": "~1.3.1"
+  },
+  "devDependencies": {
```
fzaninotto: We'll have to decide which is the default, between ng-admin only and full:

1. if ng-admin only, all dependencies below must be `dependencies`, because they are required to work with. Developers will need to include them (it can be done automatically in dev with [wiredep](https://github.com/taptapship/wiredep), that's what Yeoman does)
2. if full, all dependencies below must be `devDependencies`

I think 2 must be for testing purposes only, the target should be 1, and letting the developers do their own build phase.

grunt/grunt-requirejs.json
```diff
@@ -36,5 +36,42 @@
...
             }
         }
+    },
+    "prod-only": {
+        "options": {
+            "baseUrl": "src/javascripts",
+            "mainConfigFile": "src/javascripts/ng-admin.js",
+            "include": ["ng-admin"],
+            "exclude": [
+                "bower_components/angular-resource/angular-resource",
```
fzaninotto: How do we make sure we don't forget one (indirect) dependency?

eranimo:
This would be really nice

bower.json
manuquentin: It's the full the default one, I've named it `ng-admin-only.min.js` to not break compatibility.

jpetitcolas:
Needs rebase

fzaninotto:
Needs rebase... again. @jpetitcolas is this PR still necessary now that we aim for webpack?

fzaninotto:
Superseded by #428

Issue #332, opened 2015-02-26
Edit identity (primary key) field


vasiakorobkin:
**Given:**
* a **product** entity element with *name* custom primary key, which set as entity identifier using entity method *.identifier()*
*  Let *name* value to be *"abc"*, so we can edit this element by sending PUT request to  ***/product/abc*** URL

**Problem:**
When I try to edit *name* custom primary key setting it to, for example, *"def"*, ng-admin sends modified entity element via PUT request to ***/product/def*** URL. But no corresponding entity element does exist on this URL at the moment, so back-end fails to save changes.

**Suggested solution:**
To make ng-admin send PUT request to ***/product/abc*** URL, using the original *name* value *"abc"* even if it was modified to *"def"*. I think this should be default behavior.

fzaninotto:
Good point, that will need to be addressed some time in the future. Until we face the problem ourselves, would you like to work on a patch?

vasiakorobkin:
Well, that wasn't a part of my plans... But I'll try to spare some of my time on it.

PR #334, opened 2015-02-27
fix: Edit identity (primary key) field (#332 issue on GitHub)

vasiakorobkin:
It was a lot easier than I thought.
Fixed https://github.com/marmelab/ng-admin/issues/332
Checked and tested on example configuration.
But all changes to example configuration has been undone.
I've left only patch changes. Is that OK?

### DIFF
--- a/src/javascripts/ng-admin/Crud/form/FormController.js
+++ b/src/javascripts/ng-admin/Crud/form/FormController.js
@@ -25,6 +25,9 @@ define(function () {
         this.view = view;
         this.entity = this.view.getEntity();
 
+        // in case of entity identifier being modified
+        this.origEntityId = this.$scope.entry.values[this.entity.identifier().name()]
+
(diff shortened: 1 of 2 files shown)


src/javascripts/ng-admin/Crud/form/FormController.js
```diff
@@ -25,6 +25,9 @@ define(function () {
         this.view = view;
         this.entity = this.view.getEntity();
 
+        // in case of entity identifier being modified
+        this.origEntityId = this.$scope.entry.values[this.entity.identifier().name()]
```
fzaninotto: I'd use the simpler

```js
this.entityId = entry[this.entity.identifier().name()];
```

fzaninotto:
Almost all right. I'd just use a better name for the extra parameter - we try to avoid abbreviations in variable names.

fzaninotto:
I did the renaming myself in d02f11c. Thanks for the patch!

PR #337, opened 2015-02-27
Directly require codemirror so it optimizes into main package

easel:


fzaninotto:
I don't understand, CodeMirror already ends up into `ngadmin.min.js`, what's the purpose of this one?

easel:
That's interesting, perhaps there is a discrepancy between our versions of require.js - I have 2.1.15 and ng-admin is using 2.1.14. Typically a double-nested require won't be picked up by requirejs, and it didn't by mine. I noticed this when I loaded my site against the optimized code and most of the code mirror dependencies were flagged by google page speed as candidates for minification. Are you sure all the code mirror "helpers" are being compiled into the ng-admin.min.js?

jpetitcolas:
Very old issue. I'm closing it. Feel free to re-open it if you still encounter the issue. :)

By the way, we reviewed our compilation processes using Webpack. All required CodeMirror helpers should be compiled into the built file `ng-admin.min.js` (but not the `ng-admin-only.min.js` file).

PR #340, opened 2015-03-03
fix: 404 error for fontawesome-webfont on local example server

vasiakorobkin:
Icons from FontAwesome webfont is unavailable when running "make run" in newly clonned repo (with "make install" executed, of course). The reason of such behaviour is the absence of *assets/fonts/* folder in the example root folder */example/blog/*.
So I added "copy:assets_dev" grunt task, which copies assets folder to the example root, and included this task to the "watch:sass" task settings, where it should be IMO.


fzaninotto:
Thanks!

PR #342, opened 2015-03-03
Fix: .editable(false) on "reference" type field in EditView

vasiakorobkin:
When applying *.editable(false)* field method on field of "reference" type for EditView - we get no visible representation of this field - we get a link with empty inner text. This is because of getReadWidget() which passes an empty value to corresponding directive.
So I've slightly changed this function to fix this issue.

### DIFF
--- a/src/javascripts/ng-admin/Crud/fieldView/ReferenceFieldView.js
+++ b/src/javascripts/ng-admin/Crud/fieldView/ReferenceFieldView.js
@@ -2,7 +2,7 @@ define(function(require) {
     "use strict";
 
     function getReadWidget() {
-        return '<ma-string-column value="::entry.listValues[field.name()] || entry.values[field.name()]"></ma-string-column>';
+        return '<ma-string-column value="::entry.listValues[field.name()]"></ma-string-column>';
     }
     function getLinkWidget() {
         return '<a ng-click="gotoReference()">' + getReadWidget() + '</a>';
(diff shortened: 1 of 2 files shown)


fzaninotto:
It's not normal that you get an empty value. Does the related entity have an empty title or name?

For instance, for a Reference from Comment to Post:

```js
comment.listView().fields([
    ...,
    nga.field('post_id', 'reference')
        .label('Post')
        .targetEntity(post)
        .targetField(nga.field('title'))
])
```

The displayed value when editable is false is the post title. Your patch makes it fallback to post_id, but you shouldn't represent the related entity by a nullable field in the first place...

So either I don't understand your use case, or the patch is wrong.

vasiakorobkin:
Related entity doesn't have empty title/name.
In your example you're talking about **listView** and it seems to be no problem there.

The problem takes place in **editView**. Let's say, that I'm in need to have a Reference from Comment to Post defined only once - on Comment creation. Therefore in **editView** this Reference shouldn't be editable. 1d7f47f illustrates my attempt to get such behavior. As a result we can see an empty string instead of post title, while editing any of the Comments.

But you were right about my patch. It was wrong. c8960a4 is my another attempt to fix this issue. Now it seems to work as expected.

fzaninotto:
You patch fixes the non-editable reference in edit view, but in return breaks the editable reference...

vasiakorobkin:
Fixed in b71d686ec268287d0166565ddfe2dc8dc6bda0b8
I'll try to be more attentive next time.

fzaninotto:
Thanks!

Issue #343, opened 2015-03-03
Link to "Custom Page" in the menu


ponsfrilus:
Maybe there is already a way to do it, but if not consider it as an enhancement. It will be helpfull to link a CustomPage in the menu (https://github.com/marmelab/ng-admin/blob/master/doc/Custom-pages.md).

Thanks,
ponsfrilus

fzaninotto:
It's not possible right now, but not hard to implement. Definitely a must have.

jpetitcolas:
It is now possible, see [demo](https://github.com/marmelab/ng-admin/blob/master/examples/blog/config.js#L282-L282).

ponsfrilus:
@jpetitcolas thanks !

PR #344, opened 2015-03-03
Allow custom item in the menu

fzaninotto:
- [x] Add support for custom element (not entity) in the menu
- [ ] Mark custom menu as selected when clicked
- [ ] Add tests
- [ ] Add documentation

Refs #343


fzaninotto:
I'll use another approach.

PR #347, opened 2015-03-05
[RFR] Fix date sent to json_server in blog example

### LINKED ISSUE #345: Can't filter on date anymore with sample config
On comments list, using the default blog config:

bug-date

Date seems to use full JS format, including time. Then, as no post has been written at midnight, there is no result.

fzaninotto:
We used to send a simple string to the backend, now we send a serialized date. Unfortunately, the backend used in the blog example doesn't understand dates, and even less timezones. So we do the job in the frontend of undoing timezone offset and sending the raw date.

Closes #345

### DIFF
--- a/examples/blog/config.js
+++ b/examples/blog/config.js
@@ -162,7 +162,15 @@
                 nga.field('created_at', 'date')
                     .label('Posted')
                     .attributes({'placeholder': 'Filter by date'})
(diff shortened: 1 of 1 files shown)


fzaninotto:
I must be drunk or something: there is no such thing as a date with timezone (only a datetime with time zone). What I added as a special case should be the default case. I'll make another PR.

PR #357, opened 2015-03-16
[RFR] add an option on maFileField to recover the serverside filename

alexisjanvier:
If the uploaded file is renamed server-side, we can recover the new filename from an api return.


RobinBressan:
:+1:

jpetitcolas:
Tests are failing: same issue on master, brought by the ES6 refactoring. Can you ensure that `grunt test:local` command passes correctly?

I'm adding an issue to fix tests.

alexisjanvier:
Yes, test passes correctly on local.

PR #360, opened 2015-03-17
[RFR] Better clean of ng-admin build

jpetitcolas:
Remove `ng-configuration*` files from `build` folder, as they are already embed into `ng-admin*` files.


fzaninotto:
Thanks!

PR #364, opened 2015-03-17
support for parameterizing choice fields by the current entry

kenegozi:
This opens up contextual select tags, e.g. cascading selects.

e.g to create a cascading set of 'state' and 'city' fields, you can now do that:

     nga.field('state', 'choice').choices(states),
     nga.field('city', 'choice').choices(function(entry) {
         return cities.filter(function(city) { return city.state === entry.values.state;} );
     })

given

     var states = [ { value:'WA', label:'Washington'}, ... ];
     var cities = [ { value:'Bellevue', label: 'Bellevue', state: 'WA'} , ... ];

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,31 @@ app.config(function (NgAdminConfigurationProvider) {
         ])
         .listActions(['show', 'edit', 'delete']);
 
+   var categories = [{
+        label : 'Tech',
+        value : 'tech'
+    },{
+        label : 'Lifestyle',
+        value : 'lifestyle'
+    }];
+    var subCategories = [{
+        label : 'Computers',
+        value : 'computers',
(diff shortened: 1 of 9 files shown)


fzaninotto:
Great idea!

Although there are a couple more uses of `Field.choices`, in `ChoicesField` for instance. Could you also update these ones?

Also, would you mind adding a mention of the new syntax in the README?

kenegozi:
Good catch. I included ChoicesField, added a few tests/specs (for Choice and Choices), and a mention in the readme.

jpetitcolas:
:+1:

jpetitcolas:
Can you rebase on master, now that tests are fixed?

fzaninotto:
Superseded by #398

PR #380, opened 2015-04-02
[RFR] Batch actions config

ThieryMichel:


src/javascripts/ng-admin/Crud/misc/ViewBatchActions.js
```diff
@@ -10,7 +10,6 @@ define(function (require) {
 
         return {
             restrict: 'E',
-            transclude: true,
```
ThieryMichel: transclude was not used, and disallowed to use ng-switch on the directive

PR #381, opened 2015-04-03
[RFR] Fix exportfields config

ThieryMichel:
exportfields need to accept fields in literal form as the fields method return fields in this form.


src/javascripts/ng-admin/Crud/misc/EntryFormater.js
```diff
@@ -7,6 +7,7 @@ define(function () {
     }
 
     EntryFormatter.prototype.formatField = function formatField(field) {
+        var label = field.label() || filed.name();
```
jpetitcolas: Typo

jpetitcolas:
:+1:

PR #398, opened 2015-04-14
[RFR] Dependent choice lists

fzaninotto:
This opens up contextual select tags, e.g. cascading selects.

e.g to create a cascading set of 'state' and 'city' fields, you can now do that:

```js
 nga.field('state', 'choice').choices(states),
 nga.field('city', 'choice').choices(function(entry) {
     return cities.filter(function(city) { return city.state === entry.values.state;} );
 })
```

given

```js
 var states = [ { value:'WA', label:'Washington'}, ... ];
 var cities = [ { value:'Bellevue', label: 'Bellevue', state: 'WA'} , ... ];
```

Supersedes #364


fzaninotto:
The new e2e tests may fail because the json server used by saucelabs isn't up to date with this PR. Will be done once it is merged.

PR #410, opened 2015-04-27
[RFR] Introduce DataStore and remove entries for ReferenceField

jeromemacias:
* [x] Create DataStore to sotre all entries related to views
* [x] Remove `View.entries` property
* [x] Move `mapEntry(ies)` from View
* [x] ~~Remove Views created by code for references~~
* [x] Fix (and add ?) tests

Result of this refactoring:
* `RetrieveQueries` does not deal anymore with `Entry`
* `DataStore` now store all entries, references and filter values
* `mapEntry` is now done by `DataStore`
* `routing.js` now explicitly create and map entries
* `ReferenceField` does not extend `ChoiceField` anymore

### DIFF
--- a/src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
+++ b/src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
@@ -3,7 +3,7 @@ define(function(require) {
 
     function getReadWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}" ' +
-                 'entries="::DataStore.getEntries(field.getReferencedView())" ' +
+                 'entries="::datastore.getEntries(field.targetEntity().uniqueId + \'_list\')" ' +
                  'fields="::field.getReferencedView().fields()" ' +
                  'list-actions="::field.listActions()" ' +
                  'entity="::field.getReferencedView().entity">' +
@@ -17,7 +17,7 @@ define(function(require) {
     }
     function getWriteWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}"' +
-                  'entries="::DataStore.getEntries(field.getReferencedView())" ' +
+                  'entries="::datastore.getEntries(field.targetEntity().uniqueId + \'_list\')" ' +
                   'fields="::field.getReferencedView().fields()" ' +
                   'list-actions="::field.listActions()" ' +
                   'entity="::field.getReferencedView().entity">' +
--- a/src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
+++ b/src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
@@ -3,7 +3,7 @@ define(function(require) {
 
     function getReadWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}" ' +
-                 'entries="::field.getReferencedView().getReferencesEntries()" ' +
+                 'entries="::DataStore.getEntries(field.getReferencedView())" ' +
                  'fields="::field.getReferencedView().fields()" ' +
                  'list-actions="::field.listActions()" ' +
                  'entity="::field.getReferencedView().entity">' +
@@ -17,7 +17,7 @@ define(function(require) {
     }
     function getWriteWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}"' +
-                  'entries="::field.getReferencedView().getReferencesEntries()" ' +
+                  'entries="::DataStore.getEntries(field.getReferencedView())" ' +
                   'fields="::field.getReferencedView().fields()" ' +
                   'list-actions="::field.listActions()" ' +
                   'entity="::field.getReferencedView().entity">' +
--- a/src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
+++ b/src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
@@ -3,7 +3,7 @@ define(function(require) {
 
     function getReadWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}" ' +
-                 'entries="field.entries" ' +
+                 'entries="::field.getReferencedView().getReferencesEntries()" ' +
                  'fields="::field.getReferencedView().fields()" ' +
                  'list-actions="::field.listActions()" ' +
                  'entity="::field.getReferencedView().entity">' +
@@ -17,7 +17,7 @@ define(function(require) {
     }
     function getWriteWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}"' +
-                  'entries="field.entries" ' +
+                  'entries="::field.getReferencedView().getReferencesEntries()" ' +
                   'fields="::field.getReferencedView().fields()" ' +
                   'list-actions="::field.listActions()" ' +
                   'entity="::field.getReferencedView().entity">' +
@@ -28,5 +28,5 @@ define(function(require) {
         getLinkWidget:   getLinkWidget,
         getFilterWidget: getFilterWidget,
         getWriteWidget:  getWriteWidget,
-    }
+    };
 });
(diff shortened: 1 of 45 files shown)


src/javascripts/ng-admin/Crud/fieldView/ReferencedListFieldView.js
```diff
@@ -17,7 +17,7 @@ define(function(require) {
     }
     function getWriteWidget() {
         return '<ma-datagrid name="{{ field.getReferencedView().name() }}"' +
-                  'entries="field.entries" ' +
+                  'entries="::field.getReferencedView().getReferencesEntries()" ' +
```
fzaninotto: not sure about the `::` here

src/javascripts/ng-admin/es6/lib/DataStore/DataStore.js
```diff
@@ -0,0 +1,50 @@
+
+class DataStore {
+    constructor(name) {
+        this._name = name;
+        this._entry = new Map();
```
fzaninotto: not a big fan of the double usage. I think you should only keep the `_entries` property and related methods. For views requiring only a single entry, you'd call `store.getFirstEntry(view)`.

src/javascripts/ng-admin/es6/lib/Entity/Entity.js
```diff
@@ -13,6 +14,7 @@ import ExportView from '../View/ExportView';
 class Entity {
     constructor(name) {
         this._name = name;
+        this._dataStore = new DataStore(name);
```
fzaninotto: I don't think entities should have a datastore. Entities should be immutable (and reusable) once the configuration phase is over. Putting the DataStore there prevents this and mixes up roles.

src/javascripts/ng-admin/es6/lib/View/View.js
```diff
@@ -3,6 +3,7 @@ import Entry from "../Entry";
 class View {
     constructor(name) {
         this.entity = null;
+        this._dataStore = null;
```
fzaninotto: Same remark as for `Entity`: `View` instances should be immutable and not contain a `DataStore`. I think there should only be one single `DataStore` instance, probably linked to by the `Application` instance.

fzaninotto:
Could you add a checklist in the description to show what's left for RFR?

src/javascripts/ng-admin/Crud/repository/RetrieveQueries.js
```diff
@@ -129,40 +112,43 @@ define(function (require) {
...
         var self = this,
-            referencedValues = {},
+            referencedData = {},
             calls = [],
             singleCallFilters,
             identifiers,
             reference,
             referencedView,
-            entries,
+            datas,
```
manuquentin: `data` is already plural

src/javascripts/ng-admin/Crud/routing.js
```diff
@@ -55,17 +62,83 @@ define(function (require) {
...
+                    entries: ['dataStore', 'view', 'response', 'referencedEntries', function (dataStore, view, response, referencedEntries) {
+                        var entries = dataStore.mapEntries(
+                            view.entity.name(),
+                            view.identifier(),
+                            view.getFields(),
+                            response.data
+                        );
+
+                        // shortcut to diplay collection of entry with included referenced values
+                        dataStore.fillReferencesValuesFromCollection(entries, view.getReferences(), true);
```
jpetitcolas: Shouldn't we call this method only in the `dataStore.setEntries` or `dataStore.addEntry` methods?

src/javascripts/ng-admin/es6/lib/DataStore/DataStore.js
```diff
@@ -0,0 +1,119 @@
+import Entry from "../Entry";
+
+class DataStore {
+    constructor() {
+        this._entries = {};
```
jpetitcolas: May be a good occasion to use [ES6 Sets](http://www.nczonline.net/blog/2012/09/25/ecmascript-6-collections-part-1-sets/)?

src/javascripts/ng-admin/es6/lib/DataStore/DataStore.js
```diff
@@ -0,0 +1,119 @@
...
+
+        this._entries[name].push(entry);
+    }
+
+    getEntries(name) {
+        return this._entries[name] || [];
+    }
+
+    getFirstEntry(name) {
+        return this._entries[name][0];
```
manuquentin: Can you add a check to return the first value if the content is not empty ?

src/javascripts/ng-admin/es6/lib/DataStore/DataStore.js
fzaninotto: We tried that, unfortunately babel doesn't transpile ES6 Sets. It requires a shim and/or works only in latest Chrome versions.

src/javascripts/ng-admin/es6/lib/DataStore/DataStore.js
```diff
@@ -0,0 +1,119 @@
...
+                value: entry.values[identifier],
+                label: entry.values[name]
+            };
+        });
+    }
+
+    createEntry(entityName, identifier, fields) {
+        var entry = new Entry.mapFromRest(entityName, identifier, fields, {});
+
+        fields.forEach(function (field) {
```
manuquentin: Can you avoid `forEach` [it's 40 slower than a for loop](https://josephscott.org/archives/2014/10/javascript-performance-for-vs-foreach/)

fzaninotto: It's more like 3 times slower if you take a non-basic loop body. Using a `for` loop is a micro-optimization IMO, and makes the code less readable.

src/javascripts/ng-admin/es6/lib/Entity/Entity.js
```diff
@@ -10,9 +10,12 @@ import ShowView from '../View/ShowView';
 import BatchDeleteView from '../View/BatchDeleteView';
 import ExportView from '../View/ExportView';
 
+var index = 0;
```
jpetitcolas: IMO, it should be the `Application` responsability to keep track of its entities, and thus attributing them an unique id (via a `setId` method for instance). It would prevent a global variable.

src/javascripts/ng-admin/es6/tests/lib/DataStore/DataStore.js
```diff
@@ -0,0 +1,169 @@
...
+            { post_id: 3, title: 'How to use ng-admin', published: false}
+        ]);
+
+        assert.equal(entries.length, 3);
+        assert.equal(entries[0].identifierValue, 1);
+        assert.equal(entries[1].values.title, 'World');
+        assert.equal(entries[1].values.published, false);
+    });
+
+    it('should map some one entity when the identifier in not in the view', function () {
```
jpetitcolas: s/in not/is not/

src/javascripts/ng-admin/Crud/list/ListController.js
```diff
@@ -20,10 +20,10 @@ define(function () {
         this.loadingPage = false;
         this.filters = this.$filter('orderElement')(view.filters());
         this.hasFilters = Object.keys(this.filters).length > 0;
-        this.entries = data.entries;
+        this.dataStore = dataStore;
```
manuquentin: Should the dataStore be emptied on `$scope.$on('$destroy')` event ?

src/javascripts/ng-admin/Crud/routing.js
jeromemacias: This method is use only one time, but `setEntries` is use in multiple cases, I prefer be explicit, and call it in the routing, what do you think ?

jpetitcolas: It is called three times, but all in this routing file indeed. :)

I can't figure out any use case where we won't need to load referenced values. So, calling it externally looks dev-unfriendly to me.

jpetitcolas: My bad, I confused between `fillReferencesValuesFromCollection` and `fillReferencesValuesFromEntry`. Let it this way from now. We'll factorize later if required.

PR #441, opened 2015-05-15
[RFR] Fix empty list actions column

### LINKED ISSUE #439: Empty actions column in listView
By default, the listView shows an empty column:

image

This is, I believe, a regression: by default, the listView with not `listActions` configured used to hide the actions column.

manuquentin:
Fixes #439

### DIFF
--- a/src/javascripts/ng-admin/Crud/list/DatagridController.js
+++ b/src/javascripts/ng-admin/Crud/list/DatagridController.js
@@ -119,6 +119,11 @@ define(function () {
         this.$scope.selection = [];
     };
 
+
+    DatagridController.prototype.shouldDisplayActions = function () {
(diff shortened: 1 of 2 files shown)


src/javascripts/ng-admin/Crud/list/DatagridController.js
```diff
@@ -119,6 +119,11 @@ define(function () {
         this.$scope.selection = [];
     };
 
+
+    DatagridController.prototype.shouldDisplayActions = function () {
```
fzaninotto: Make it a value in the scope instead, this cannot change

manuquentin: Done

Issue #447, opened 2015-05-19
Entity without listView still shows list buttons


fzaninotto:
Use case: I define an entity with only a showView, accessible from a related entity (e.g. comments, accessible only from the commented post). On purpose, I decide to not define a listView. 

The problem is that the showView still contains a "List" action button, leading to an empty list.

This is a use case we must support.

manuquentin:
Views in ng-admin are [enabled by default](https://github.com/marmelab/ng-admin/blob/a312bdc3a1d2505dc403449c36597a8e1f78099f/src/javascripts/ng-admin/es6/lib/View/View.js#L11).

You have to disable them manually.

We can disable them by default & enable them when using customization methods like `fields()`,  `title()`.
But if someone doesn't customize the deletionView for instance, the view won't be enabled.

PR #449, opened 2015-05-20
[RFR] Fix button display when a view is disabled

manuquentin:
Refs #447 

- [x] Add missing condition to hide buttons
- [x] Disable views by default (except deleteView)
- [x] Use the number of fields to know if a view is enabled


jeromemacias:
It will be a BC break, no ?

manuquentin:
Not really, it break things that with weird behaviour.
Like `ListView` without fields of `MenuView` for an entity without `ListView`.

Issue #362, opened 2015-03-17
Force HTTP Method


paulogr:
I know that it's not right but can I force a POST method in the listView type ?

I got a API that only works with POST method and I need to force all requests to POST.

Thanks
Paulo

fzaninotto:
Hi Paulo,

You can't for now - Restangular interceptors allow you to override many aspects of an HTTP request, but not the method. A similar concerned is being addressed in #286, feel free to work on something on your side and send us a pull request if you succeed.

paulogr:
Ok man, thanks.

fzaninotto:
#555 tries to do what you ask

PR #451, opened 2015-05-21
added custom HTTP-Methods for create and update

matsluni:
Hi,

for a current task I need the possibility to set a custom HTTP method for a create operation, in my case 'put' instead of 'post'. I saw a similar request in #362.  

I gave it a try in this pull request.  The change I did works for me, so please consider merging if you think this is a useful addition.

If you think another approach for the same goal is better just get back to me.
I also tried to add tests for this in CreateQueriesSpec.js and UpdateQueriesSpec.js but couldn't figure out to set the xxxMethod on the entity. So, if you could give me a hint I will add this as well.

Kind regards,
Matthias

### DIFF
--- a/src/javascripts/ng-admin/Crud/repository/CreateQueries.js
+++ b/src/javascripts/ng-admin/Crud/repository/CreateQueries.js
@@ -26,8 +26,8 @@ define(function (require) {
     CreateQueries.prototype.createOne = function (view, rawEntity) {
 
         var method = view.entity.createMethod(),
-            url = this.Restangular.oneUrl(view.entity.name(), this.config.getRouteFor(view.entity, view.getUrl(), view.type)),
-            operation = method ? url.customOperation(method, null, {}, {}, rawEntity) : url.customPOST(rawEntity);
+            resource = this.Restangular.oneUrl(view.entity.name(), this.config.getRouteFor(view.entity, view.getUrl(), view.type)),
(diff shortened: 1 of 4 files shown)


src/javascripts/ng-admin/Crud/repository/CreateQueries.js
```diff
@@ -24,12 +24,14 @@ define(function (require) {
...
     CreateQueries.prototype.createOne = function (view, rawEntity) {
-        return this.Restangular
-            .oneUrl(view.entity.name(), this.config.getRouteFor(view.entity, view.getUrl(), view.type))
-            .customPOST(rawEntity)
-            .then(function (response) {
-                return response.data;
-            });
+
+        var method = view.entity.createMethod(),
+            url = this.Restangular.oneUrl(view.entity.name(), this.config.getRouteFor(view.entity, view.getUrl(), view.type)),
```
fzaninotto: The variable name isn't well chosen. Use `resource` instead. Same in `UpdateQueries`

matsluni:
Ok, I changed the variable name to resource. Is there anything more I can do?

manuquentin:
Can you rebase please ?

matsluni:
I tried to rebase and saw that you moved some pieces around! ;-) I have look closer to rebase correctly.

jeromemacias:
You will need to do all this modifications on the new https://github.com/marmelab/admin-config repository. Sorry for inconvenience.

PR #456, opened 2015-05-22
[RFR] Update README for Webpack

jpetitcolas:


manuquentin:
Tests fails :stuck_out_tongue_closed_eyes:

fzaninotto:
If I understand correctly, "standalone" contains all dependencies and "" contains only the ng-admin code. It's kind of backwards, and breaks BC badly...

jpetitcolas:
True for BC. Let's switch to `ng-admin-only` and `ng-admin` in this case.

fzaninotto:
or `ng-admin-no-deps` to make it obvious.

fzaninotto:
wait, it's not obvious. Let's keep `only`.

PR #457, opened 2015-05-22
[RFR] Use admin-config files

manuquentin:


src/javascripts/ng-admin.js
```diff
@@ -1,7 +1,7 @@
 require('./ng-admin/Main/MainModule');
 require('./ng-admin/Crud/CrudModule');
 
-var Factory = require('./ng-admin/es6/lib/Factory');
+import Factory from 'admin-config/lib/Factory';
```
jeromemacias: You should use `require` as no `import` syntax is use in the project (even if it's work).

manuquentin: Done

jpetitcolas:
:+1:

PR #461, opened 2015-05-22
[RFR] Fix maExportToCsvButton does not return entries references and empty data when filters contains boolean field

jeromemacias:


src/javascripts/ng-admin/Crud/list/ListController.js
```diff
@@ -17,6 +17,7 @@ define(function () {
         this.actions = view.actions();
         this.batchActions = view.batchActions();
         this.loadingPage = false;
+        this.search = $stateParams.search;
```
jeromemacias: This `$stateParams.search` is ok.

src/javascripts/ng-admin/Crud/button/maExportToCsvButton.js
```diff
@@ -25,27 +26,78 @@ define(function () {
...
-                        var csv = Papa.unparse(results);
-                        var fakeLink = document.createElement('a');
-
-                        fakeLink.setAttribute('href', 'data:application/octet-stream;charset=utf-8,' + encodeURIComponent(csv));
-                        fakeLink.setAttribute('download', scope.entity.name() + '.csv');
-                        fakeLink.click();
-                    }, function (error) {
-                        notification.log(error.message, {addnCls: 'humane-flatty-error'});
-                    });
+                    ReadQueries.getAll(exportView, -1, scope.search(), $stateParams.sortField, $stateParams.sortDir)
```
jeromemacias: If we use `$stateParams.search` instead of `scope.search()`, it's not the right value (but it should be the same value as in `ListController`...).

src/javascripts/ng-admin/Crud/misc/ViewActions.js
```diff
@@ -17,7 +17,8 @@ define(function (require) {
                 'entity': '=',
                 'selection': '=',
                 batchButtons: '&',
-                datastore: '='
+                datastore: '=',
+                search: '='
```
jpetitcolas: Not used

jeromemacias: It is... Angularjs black magic!

PR #479, opened 2015-05-29
[WIP] Fix tests due to Angular 1.4 version switch

jpetitcolas:
Still WIP as I'd like to ensure nothing is broken manually first.

For more details about this change: https://docs.angularjs.org/guide/migration#ngoptions

### DIFF
--- a/src/javascripts/test/karma.conf.js
+++ b/src/javascripts/test/karma.conf.js
@@ -13,8 +13,8 @@ module.exports = function (config) {
         frameworks: ['jasmine'],
         files: [
             '../../node_modules/angular/angular.js',
-            '../../node_modules/angular-bootstrap/dist/ui-bootstrap.min.js',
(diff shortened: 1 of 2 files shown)


jeromemacias:
Test e2e failed.

fzaninotto:
Can't we support both 1.3 and 1.4?

jeromemacias:
We maybe need to provide 2 separate packages for the full ng-admin (the one ith AngularJs included).

Something like `ng-1.4-admin.min.js` and `ng-admin.js`.
It could be possible to have an `ng-admin-only.js`with an angular 1.3 and 1.4 compatibility.

jpetitcolas:
Not a big fan of having to maintain several versions. It would be better to continue supporting 1.3 while it is stable. When 1.4 will be stable, then deliver a new release using the new version.

fzaninotto:
1.4 is stable. We should identify the parts in the codebase that break on 1.4.

jpetitcolas:
Soon superseeded by #567.

PR #480, opened 2015-05-29
[RFR] fix pagination

ThieryMichel:


src/javascripts/ng-admin/Crud/list/ListController.js
```diff
@@ -47,9 +47,16 @@ define(function () {
 
         this.ReadQueries
             .getAll(this.view, page, true, this.search, this.sortField, this.sortDir)
-            .then(function (nextData) {
```
jeromemacias: You also need to fix parameters passed to `getAll`method. The third one (`true`) does not exists anymore.

src/javascripts/ng-admin/Crud/list/ListController.js
```diff
@@ -47,9 +47,16 @@ define(function () {
...
+            .then(function (response) {
                 progression.done();
-                self.entries = self.entries.concat(nextData.entries);
+                self.dataStore.mapEntries(
```
jeromemacias: You need to retrieve references before mapping and set entries to the datastore.

src/javascripts/ng-admin/Crud/list/ListController.js
```diff
@@ -46,10 +46,21 @@ define(function () {
         progression.start();
 
         this.ReadQueries
-            .getAll(this.view, page, true, this.search, this.sortField, this.sortDir)
-            .then(function (nextData) {
+            .getAll(this.view, page, this.search, this.sortField, this.sortDir)
+            .then(function (response) {
                 progression.done();
-                self.entries = self.entries.concat(nextData.entries);
+                var references = self.view.getReferences();
```
jeromemacias: You also need to retrieve references and store them into the datastore.

See https://github.com/marmelab/ng-admin/blob/a7cc56da34ed9f249df003128594124e9b2b632e/src/javascripts/ng-admin/Crud/routing.js#L73-L118.

ThieryMichel: Are you sure all of this is necessary. Because, on my project, the references were correctly retrieved, even the one that were not on the first page.
So it seems to me that the code in routing.js does the job, and don't need to be repeated.

PR #483, opened 2015-05-30
[RFR] Stay on Angularjs 1.3 version

jeromemacias:
I suggest to stay on AngularJS 1.3 as 1.4 breaks many things in ng-admin core and project using it (custom directive for example).
Because ng-admin embed by default the AngularJS dependency, all projects will be updated to the 1.4 version, and this will probably broke something.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "devDependencies": {
     "admin-config": "^0.1.3",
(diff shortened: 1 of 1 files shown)


manuquentin:
What do you think @jpetitcolas ?

jpetitcolas:
Fine for me.

PR #491, opened 2015-06-03
added userdefined create and update HTTP Method

matsluni:
Hi,

this is the complement to marmelab/admin-config#8 and the 2nd approach for #451.
Let me know what you think.

Mats


fzaninotto:
Great addition, thanks!

PR #508, opened 2015-06-15
[RFR] Send API request on reference auto-complete

jpetitcolas:
Instead of having a pre-define set of choices in reference fields, send API requests at will to fetch related records.

anim

**Left tasks:**

- [x] Implement API auto-complete on single reference fields (post id)
- [x] Display correctly selected value
- [x] Implement API auto-complete on multiple reference fields (tags)
- [x] Implement tests
- [x] If no fields configured, disable refresh and populate with all data (for BC)
- [x] Update admin-config repository to deal with `refreshDelay`
- [x] Update documentation
- [x] Update blog sample configuration
- [x] Test it on real world project


src/javascripts/ng-admin/Crud/field/maChoiceField.js
```diff
@@ -14,10 +14,35 @@ function maChoiceField($compile) {
...
+                        if (!search) {
+                            return;
+                        }
+
+                        var referenceFields = {};
+                        referenceFields[scope.name] = field;
+
+                        return ReadQueries.getAllReferencedData(referenceFields, search)
+                            .then(r => r[field.name()])
+                            .then((results) => {
```
fzaninotto: You don't need parentheses for a single argument

src/javascripts/ng-admin/Crud/field/maChoiceField.js
```diff
@@ -14,10 +14,35 @@ function maChoiceField($compile) {
...
+                                return results.map(function(r) {
+                                    return {
+                                        value: r.id,
```
fzaninotto: should not be hardcoded

src/javascripts/test/karma.conf.js
```diff
@@ -12,7 +12,7 @@ module.exports = function (config) {
         browsers: [process.env.CI ? 'PhantomJS' : 'Chrome'],
         frameworks: ['jasmine'],
         files: [
-            '../../node_modules/angular/angular.min.js',
+            '../../node_modules/angular/angular.js',
```
jpetitcolas: Revert

src/javascripts/test/unit/Crud/field/maReferenceFieldSpec.js
```diff
@@ -0,0 +1,98 @@
+xdescribe('ReferenceField', function() {
```
jpetitcolas: Skipped as directive doesn't want to compile.

jpetitcolas:
Switching to RFR. I still need to test it on a real world project before merging.

jpetitcolas:
Tested on real world project. Seems to work. Removing the [DO NOT MERGE].

src/javascripts/ng-admin/Crud/field/maChoiceField.js
```diff
@@ -17,18 +17,13 @@ function maChoiceField($compile) {
...
 
-                    var choices;
-                    if (field.type() === 'reference' || field.type() === 'reference_many') {
-                        choices = scope.datastore().getChoices(field);
-                    } else {
-                        choices = field.choices();
-                    }
-                    scope.getChoices = typeof(choices) === 'function' ? choices : function() { return choices; };
+                    var choices = field.choices();
+                    scope.choices = typeof(choices) === 'function' ? choices(scope.entry) : choices;
```
fzaninotto: I'm surprised this works, I had to return a function because Angular complained about an infinite look. Needs further testing.

src/javascripts/ng-admin/Crud/field/maReferenceField.js
```diff
@@ -0,0 +1,87 @@
...
+                                scope.choices = formattedResults;
+                                scope.$root.$$phase || scope.$apply();
+                            });
+                    };
+
+                    scope.refreshDelay = field.refreshDelay();
+
+                    var refreshAttributes = scope.refreshDelay !== null ? 'refresh-delay="{{ refreshDelay }}" refresh="refreshChoices($select.search)"' : '';
+
+                    var template = `
```
fzaninotto: I don't understand why you repeat the `<ma-choice-field>` code instead of using the `<ma-choice-field>` directive directly.

src/javascripts/ng-admin/Crud/field/maReferenceField.js
```diff
@@ -0,0 +1,87 @@
...
+        scope: {
+            'field': '&',
+            'value': '=',
+            'entry':  '=?',
+            'datastore': '&?'
+        },
+        restrict: 'E',
+        compile: function() {
+            return {
+                pre: function(scope, element) {
```
fzaninotto: I didn't pay attention when you updated `<ma-choice-field>`, but why do you need this instead of a simple `link` function?

src/javascripts/ng-admin/Crud/field/maReferenceField.js
```diff
@@ -0,0 +1,87 @@
...
+
+                    var select = element.children()[0];
+                    var attributes = field.attributes();
+                    for (var name in attributes) {
+                        select.setAttribute(name, attributes[name]);
+                    }
+
+                    // Pre-fill component with given value if any
+                    if (scope.refreshDelay !== null && scope.value) {
+                        ReadQueries.getOne(field.targetEntity(), null, scope.value)
```
fzaninotto: I don't understand, the API fetching of one to many reference is already done in the routing. Why repeat it here?

src/javascripts/ng-admin/Crud/field/maReferenceManyField.js
```diff
@@ -0,0 +1,108 @@
...
+                    scope.choices = [];
+
+                    var valueFieldName = field.targetEntity().identifier().name();
+                    var labelFieldName = field.targetField().name();
+
+                    scope.refreshChoices = function(search) {
+                        var referenceFields = {};
+                        referenceFields[scope.name] = field;
+
+                        return ReadQueries.getAllReferencedData(referenceFields, search)
```
fzaninotto: Same remark as above: this is already done in the routing `resolve` phase, you're fetching twice the same data.

fzaninotto:
If you have to put some API fetching logic in a directive, please move it to a service.

src/javascripts/test/unit/Crud/field/maReferenceFieldSpec.js
```diff
@@ -0,0 +1,123 @@
...
+
+    it('should be an ui-select field', function() {
+        var element = $compile(directiveUsage)(scope);
+        scope.$digest();
+
+        var uiSelect = element[0].querySelector('.ui-select-container');
+        expect(uiSelect).toBeTruthy();
+    });
+
+    it('should call remote API when inputting first characters', function () {
```
fzaninotto: Something bothers me: users can't opt out for ajax autocompletion. That means that if their API doesn't implement the search query param, reference fields cease to work. This is a serious BC break. 

Autocomplete should be opt-in ; the default behavior should still be to fetch all possible values, and only autocomplete locally.

jpetitcolas: Actually, this is already dealt through the `field.refreshDelay` parameter. If it's null, then [all choices would be retrieved](https://github.com/marmelab/ng-admin/pull/508/files#diff-dd582c1c352da5fbab2342624584018dR123), the same as before my changes.

I put the API request behavior by default, as it sounds more logical to me for references. Indeed, user doesn't know how many records there is remotely. And it is more probable that there is a lot. It would then be limited by the `perPage` parameter. 

This is indeed a BC break, but it improves the component behavior IMO.

jpetitcolas: As seen with @fzaninotto, we'll set `refreshDelay` to null by default.

src/javascripts/ng-admin/Crud/field/maReferenceField.js
jpetitcolas: This is required, as the `link` function would apply on the compiled `ui-select` directive. Yet, I want the custom attributes to be set on the `ui-select` raw element to be sure it propagates them correctly, depending of its custom behavior. For instance, where should we set the `required` attribute on the final ui-select template?

Using `compile` allows to set attributes on the ui-select.

src/javascripts/test/unit/Crud/field/maReferenceFieldSpec.js
fzaninotto: Moreover, you must make it more explicit that in the doc that a Reference is populated by the entire list of references, and that if the developer wants to use autocomplete, they should set the refreshdelay to something more than null.

src/javascripts/ng-admin/Crud/field/maReferenceField.js
fzaninotto: Gotcha

src/javascripts/ng-admin/Crud/field/maReferenceManyField.js
jpetitcolas: We need it here as we have to refresh data according to user input. Moving it to a dedicated service.

jpetitcolas:
Refresh logic moved to a dedicated service. Ready for another review.

jeromemacias:
Is it possible to disable first request (in routing) for retrieving references values in case of `refreshDelay` is not null ?
Something like `view.getReferencesWithNoRefreshDelay()`.

jpetitcolas:
Back to WIP for optimizing number of sent requests and to fix autocomplete, which seems to be made only on local records.

jpetitcolas:
Ready for review. 

Failing tests are due to the [admin-config PR](https://github.com/marmelab/admin-config/pull/19).

jeromemacias:
:+1:

fzaninotto:
Great work, thanks a lot!

endel:
Hey guys!

Is there any way to get the `Field` reference inside the `filters` function?
It seems that the scope gets complete lost when the function is called. 

Check out this example:

```
let field = nga.field("auth_id", 'reference').
  targetEntity(entities[plural]).
  targetField(nga.field(label_field)).
  filters(function(search) {
    console.log(field) // <= Uncaught ReferenceError: field is not defined
  }).
  remoteComplete(true, { refreshDelay: 300 }).
  singleApiCall(aggregateIds);
```

What could I do?

fzaninotto:
@endel: not for now. But this `reference.filters()` method needs an overhaul anyway.

fzaninotto:
`reference.filters()` has just been renamed to `reference.permanentFilters()` (refs #588).

@endel I can't reproduce your issue. Could it be due to babel translation? What if you use a `var` instead?

PR #509, opened 2015-06-15
[RFR] Introduce new way to define dashboard datasources

fzaninotto:
Depends on https://github.com/marmelab/admin-config/pull/14

```js
// customize dashboard
admin.dashboard(nga.dashboard()
    .addCollection('posts', nga.collection(post)
        .title('Recent posts')
        .perPage(5) // limit the panel to the 5 latest posts
        .fields([
            nga.field('title').isDetailLink(true).map(truncate)
        ])
        .order(1)
    )
    .addCollection('comments', nga.collection(comment)
        .title('Last comments')
        .perPage(5)
        .fields([
            nga.field('id'),
            nga.field('body', 'wysiwyg')
                .label('Comment')
                .stripTags(true)
                .map(truncate),
            nga.field(null, 'template') // template fields don't need a name in dashboard view
                .label('')
                .template('<post-link entry="entry"></post-link>') // you can use custom directives, too
        ])
        .order(2)
    )
    .addCollection('tags', nga.collection(tag)
        .title('Recent tags')
        .perPage(10)
        .fields([
            nga.field('id'),
            nga.field('name'),
            nga.field('published', 'boolean').label('Is published ?')
        ])
        .order(3)
    )
);
```

- [x] add support for new `nga.dashboard()` syntax
- [x] Add compatibility layer to make old `dashboardView()` work
- [x] clean up dashboard view
- [x] rename `PanelBuilder`
- [x] document new syntax
- [x] Add example of custom dashboard
- [x] Fix tests


README.md
```diff
@@ -793,6 +781,34 @@ Define a function that returns parameters for filtering API calls. You can use i
                 })
         ]);
 
+## Customizing the API Mapping
+
+All HTTP requests made by ng-admin to your REST API are carried out by [Restangular](https://github.com/mgonto/restangular), which is like `$resource` on steroids.
+
+The REST specification doesn't provide enough detail to cover all requirements of an administration GUI. ng-admin makes some assumptions about how your API is designed. All of these assumptions can be overridden by way of [Restangular's request and response interceptors](https://github.com/mgonto/restangular#addresponseinterceptor).
```
jpetitcolas: I would also add a few words about [Restangular element transformers](https://github.com/mgonto/restangular#addelementtransformer).

UPGRADE-0.8.md
```diff
@@ -20,3 +20,56 @@ The `Field.identifier(true)` does not exist anymore. Instead, you must specify t
 + entity.identifier(nga.field('id'))
 ```
 
+## `dashboardView()` is deprecated, use `admin.dashboard()` instead
+
+To let developers customize the admin dashboard at will, ng-admin 0.8 decouples the dashboard data and presentation. You can setup the dashboard datasources and template on a new member of ythe admin class, `dahboard()`:
```
jpetitcolas: Typo: `ythe`

jpetitcolas: And `dahboard`

doc/Dashboard.md
```diff
@@ -0,0 +1,159 @@
...
+
+```js
+admin.dashboard(nga.dashboard()
+    .template(templateString)
+);
+```
+
+Here is a copy of the default dashboard template, which you can use as a starting point for a custom dashboard template:
+
+```html
```
jpetitcolas: Can't we just add a link to the dashboard template? I'm affraid this code would be desynchronized over time.

jpetitcolas:
Fine for me, even if there may be a confusion between the `collection` name, which is sometimes used as a `listView` in the `ReadQueries.getAll` method for instance. We should consider (probably in another PR) to homogenize these names.

doc/Dashboard.md
fzaninotto: True, but the reading experience would be worse. It's our responsibility to keep code, tests, and documentation on the same track.

fzaninotto:
I'm also not very happy about the addition of a new concept, but this one goes in the right direction (separating data from presentation). `listView` mixes three concepts: the definition of the collection to fetch (fields, per page, etc.), the presentation (title, filters, etc), and the features (filters, actions). As it's all-or-nothing, I preferred to introduce this collection concept, which is what should be used in every ReadQuery instead of view. But that's, as you said, for another refactoring.

README.md
fzaninotto: I refer to the extensive documentation chapter about API Mapping two lines down, I think it's enough.

fzaninotto:
Green!

fzaninotto:
Back to WIP, I discovered a regression in dashboard panels title

fzaninotto:
And we're back to RFR.

PR #511, opened 2015-06-16
[RFR] Set default format in date fields

manuquentin:
Depends on [admin-config#15](https://github.com/marmelab/admin-config/pull/15)

Date format can't be shared with other projects as the formatting is done by [angular](https://docs.angularjs.org/api/ng/filter/date).


jeromemacias:
PR on ng-admin had been merged, can you update this PR ?

PR #528, opened 2015-06-24
[RFR] Fix default date format in date field

### LINKED ISSUE #520: Formatting of date fields not working as expected
I fail to to change the default format of a date field. My defaultValue is set to ```new Date()```. When the page has loaded, this is displayed as:
```
Fri Jun 19 2015 09:00:01 GMT+0200 (CEST)
```
As soon as I select a date using the date picker, the format is correctly changed to the correct (default) format (```yyyy-MM-dd```):
```
2015-06-19
```
This can easily be reproduced in the demo application at
http://ng-admin.marmelab.com/#/comments/create

When I try to set the defaultValue to something like ```2015-06-19``` (which is probably wrong because it's just a string now) then at least the value is displayed correctly. However, when the form is submitted, then an empty value is sent (it's empty even if the field is marked as required):
```
    nga.field('startdate', 'date')
        .format('yyyy-MM-dd')
        .defaultValue('2015-06-19')
        .validation({ required: true });
```

What is the right way to change the default format and value of a date field?

jpetitcolas:
Fixes #520 ([known issue on Angular Bootstrap](https://github.com/angular-ui/bootstrap/issues/2659)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "admin-config": "^0.2.1",
     "angular": "~1.3.15",
     "angular-bootstrap": "^0.12.0",
-    "angular-mocks": "^1.3.15",
+    "angular-mocks": "1.3.14",
     "angular-numeraljs": "^1.1.6",
     "angular-sanitize": "^1.3.15",
     "angular-ui-codemirror": "^0.3.0",
--- a/src/javascripts/ng-admin/Crud/field/maDateField.js
+++ b/src/javascripts/ng-admin/Crud/field/maDateField.js
@@ -1,58 +1,55 @@
-/*global define*/
+/**
+ * Edition field for a date - a text input with a datepicker.
+ *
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "admin-config": "^0.2.1",
     "angular": "~1.3.15",
     "angular-bootstrap": "^0.12.0",
-    "angular-mocks": "^1.3.15",
+    "angular-mocks": "1.3.14",
```
jpetitcolas: To prevent from a `$browser.cookies is undefined` introduced in 1.3.15.

src/javascripts/ng-admin/Crud/field/maDateField.js
```diff
@@ -1,58 +1,55 @@
-/*global define*/
```
jpetitcolas: Nothing changed here, just flatten the file to look more ES6-ish.

PR #538, opened 2015-06-30
[RFR] Delete view: back to previous state, not edit or list view

vasiakorobkin:
If I press `Cancel` in *delete view* I want to go back where I came from, not to the deleted object's *edit view* (the delete button may be pressed in *list view* or in `referenced_list` field of other entity, not always in *edit view*; also *list view* or *edit view* or both may be disabled for deleted entity - than we would have an error). Same thing with after-delete-redirect.
I think it is much more logical to just redirect user to the previous entry in history, than to specifically *edit view* or *list view*. Is there some problems with this approach? I don't see any.

### DIFF
--- a/src/javascripts/ng-admin/Crud/delete/DeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/DeleteController.js
@@ -3,7 +3,7 @@
 define(function () {
     'use strict';
 
-    var DeleteController = function ($scope, $state, WriteQueries, notification, params, view, entry) {
+    var DeleteController = function ($scope, $window, $state, WriteQueries, notification, params, view, entry) {
         this.$scope = $scope;
         this.$state = $state;
         this.WriteQueries = WriteQueries;
@@ -17,6 +17,7 @@ define(function () {
         this.notification = notification;
         this.$scope.entry = entry;
         this.$scope.view = view;
+        this.goBack = function(){ $window.history.back(); }
         $scope.$on('$destroy', this.destroy.bind(this));
     };
 
@@ -26,10 +27,7 @@ define(function () {
 
         this.WriteQueries.deleteOne(this.view, this.entityId).then(function () {
 
-            $state.go($state.get('list'), angular.extend({
-                entity: entityName,
-                id: this.entityId
-            }, $state.params));
+            this.goBack();
             notification.log('Element successfully deleted.', { addnCls: 'humane-flatty-success' });
         }.bind(this), function (response) {
             // @TODO: share this method when splitting controllers
@@ -43,12 +41,7 @@ define(function () {
     };
 
     DeleteController.prototype.back = function () {
-        var $state = this.$state;
-
-        $state.go($state.get('edit'), angular.extend({
-            entity: this.entity.name(),
-            id: this.entityId
-        }, $state.params));
+        this.goBack();
     };
 
     DeleteController.prototype.destroy = function () {
@@ -59,7 +52,7 @@ define(function () {
         this.entity = undefined;
     };
 
-    DeleteController.$inject = ['$scope', '$state', 'WriteQueries', 'notification', 'params', 'view', 'entry'];
+    DeleteController.$inject = ['$scope', '$window', '$state', 'WriteQueries', 'notification', 'params', 'view', 'entry'];
 
     return DeleteController;
 });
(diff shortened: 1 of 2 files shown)


src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -3,7 +3,7 @@
 define(function () {
     'use strict';
 
-    var DeleteController = function ($scope, $state, WriteQueries, notification, params, view, entry) {
+    var DeleteController = function ($scope, $window, $state, WriteQueries, notification, params, view, entry) {
```
fzaninotto: `$state` isn't used anymore. Please remove it entirely from the controller.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -17,6 +17,7 @@ define(function () {
         this.notification = notification;
         this.$scope.entry = entry;
         this.$scope.view = view;
+        this.goBack = function(){ $window.history.back(); }
```
fzaninotto: please make it a property of the controller prototype instead, like other functions.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -43,12 +41,7 @@ define(function () {
     };
 
     DeleteController.prototype.back = function () {
```
fzaninotto: Why not implement the "back" logic in this method instead?

fzaninotto:
I remember thinking about it, too, but I think this creates problems in e2e tests. As protractor failed to run, I relaunched the build, let's see.

vasiakorobkin:
Local version of Protractor reported no failures now.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -43,23 +40,19 @@ define(function () {
...
-        var $state = this.$state;
+        var $window = this.$window;
```
jpetitcolas: No need of a temp variable here. Just do a `this.$window.history.back()`.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -22,14 +22,11 @@ define(function () {
...
+            entityName = this.entity.name(),
+            $window = this.$window;
 
         this.WriteQueries.deleteOne(this.view, this.entityId).then(function () {
-
-            $state.go($state.get('list'), angular.extend({
-                entity: entityName,
-                id: this.entityId
-            }, $state.params));
+            $window.history.back();
```
jpetitcolas: Call `this.back()` instead.

jpetitcolas:
This PR needs rebase. :)

fzaninotto:
Bump, can you rebase and refactor the `back()` method ?

vasiakorobkin:
Rebased branch and `git push -f`'ed it to `vasiakorobkin/ng-admin`. One test is failing, but it is related to filter not reappearing, so I don't think that it related to my pull requrest.

jpetitcolas:
Tests on master are passing. Can you try to rebase to check if it fixes your PR?

vasiakorobkin:
Rebased. All test has been passed successfully.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -22,14 +22,11 @@ define(function () {
...
+            entityName = this.entity.name(),
+            $window = this.$window;
 
         this.WriteQueries.deleteOne(this.view, this.entityId).then(function () {
-
-            $state.go($state.get('list'), angular.extend({
-                entity: entityName,
-                id: this.entityId
-            }, $state.params));
+            this.back();
```
ahgittin: looks to me like this causes a regression:  if you click "delete" while on the `showView` page, the new code sends you "back" to that page which 404's.

fzaninotto: @ahgittin this is taken care of in #642. Could you test it and report your findings in the PR?

jpetitcolas: @ahgittin: incoming #642 indeed fixes this issue. If you are on the show view and click delete, the controller is going to check if the previous page is related to the deleted entity. If so, it redirects you to the list view. Otherwise, to the previous page.

PR #542, opened 2015-07-01
Add a configuration for csv export

AdJez:
Hi !
First, thanks a lot for this project it's nicely written and saves a lot of time!

I have just added a little provider for csv exportation. I thought it might be useful for others. Tell me what you think about it !

Adrien Jézégou

### DIFF
--- a/src/javascripts/ng-admin/Crud/button/CsvExportConfiguration.js
+++ b/src/javascripts/ng-admin/Crud/button/CsvExportConfiguration.js
@@ -1,0 +1,26 @@
+define(function () {
+    'use strict';
+    function CsvExportConfiguration() {
+        this.csvExportConfig = {
+            papaparseConfig : {
+                quotes: false,
+                delimiter: ",",
+                newline: "\r\n"
+            },
+            getPapaparseConfig: function() {
+                return this.papaparseConfig;
+            }
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


src/javascripts/ng-admin/Crud/button/CsvExportConfiguration.js
```diff
@@ -0,0 +1,26 @@
+define(function () {
```
fzaninotto: Why put that under the button dir?

fzaninotto: We're not using requirejs anymore, but webpack. Please leave the `define` aside and use `export` instead.

fzaninotto:
Great idea! However, I think the implementation should be angular-independant (and therefore made inside `admin-config` instead of `ng-admin`). We're trying to abstract the implementation away from the configuration, which allows us to implement [react-admin](https://github.com/marmelab/react-admin) using the exact same configuration as ng-admin.

Would you like to try and make the changes in `admin-config` instead?

AdJez:
Thanks for this quick response, i'll make this changes into admin-config instead of ng-admin and send you once done.

jpetitcolas:
Any news on it? :)

fzaninotto:
Superseded by #933

PR #546, opened 2015-07-03
[WIP] Move ng-admin as a dedicated directive

jpetitcolas:
First step towards a full directive conversion.

- [x] Create a `<ma-ng-admin />` directive
- [ ] Add a backward compatibility layer and mark [NgAdminConfiguration](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Main/component/provider/NgAdminConfiguration.js) as deprecated
- [ ] Allow to configure admin without transpiler
- [ ] Update documentation


src/javascripts/ng-admin/Main/component/directive/maNgAdmin.js
```diff
@@ -0,0 +1,18 @@
+function maNgAdminDirective(NgAdminConfiguration) {
+    'use strict';
+
+    return {
+        scope: {
+            'configuration': '&'
```
fzaninotto: Well, it's not used at all since you override it with the `NgAdminConfiguration` value. And that's a problem. I would very much like to write

```html
<ma-ng-admin configuration="myCustomValue"></ma-ng-admin>
```

But even after modifying line 10 to use `configuration` from scope as default value, it doesn't work because you depend on the hardcoded `NgAdminConfiguration` value everywhere in the code (and especially in routing).

You have to find a way for this to work without any global value, just using the current scope. That means that the initial state should depend on the current scope.

Note: I've been looking for a solution for this problem for the past hour and couldn't find one. It's not an easy problem. Good luck.

fzaninotto: If it's hard, it may be because it's not possible, see http://stackoverflow.com/questions/22313492/angular-ui-router-get-two-states-active-in-parallel. Bummer.

fzaninotto: I somehow managed to pass the configuration from the new directive to controllers down the ui-view tree. 



The `<ma-ng-admin>` directive controller activates a new state called "initialize", passing it the configuration from the scope as view parameter:

```js
function maNgAdminDirective(NgAdminConfiguration, $state) {
    'use strict';

    return {
        scope: {
            'configuration': '&'
        },
        restrict: 'E',
        link: function(scope) {
            $state.go('initialize', { configuration: scope.configuration() || NgAdminConfiguration })
        },
        template: `<div ui-view></div>`
    };
}

maNgAdminDirective.$inject = ['NgAdminConfiguration', '$state'];

export default maNgAdminDirective;
```

Here is the new state definition:

```js
    $stateProvider.state('main', {
        abstract: true,
        templateProvider: ['NgAdminConfiguration', function(configuration) {
            return configuration.layout() || layoutTemplate;
        }],
        data: {
            adminConfig: {}
        }
    });
    $stateProvider.state('initialize', {
        parent: 'main',
        params: {
            configuration: null
        },
        controller: ['$state', '$stateParams', function($state, $stateParams) {
            $state.current.data.adminConfig = $stateParams.configuration;
            $state.go('dashboard');
        }]
    });
```

The only purpose of this `initialize` state is to populate the state `data` (which is declared on the `main` state, and therefore accessible to all child states`). So now the configuration is accessible to all child view:

```js
    $stateProvider.state('dashboard', {
        parent: 'main',
        url: '/dashboard?sortField&sortDir',
        params: {
            sortField: null,
            sortDir: null
        },
        controller: 'DashboardController',
        controllerAs: 'dashboardController',
        templateProvider: function() {
            return adminConfig.dashboard().template() || dashboardTemplate;
        },
        resolve: {
            dataStore: dataStoreProvider(),
            adminConfig: ['$state', function($state) {
                return $state.current.data.adminConfig;
            }],
            hasEntities: ['adminConfig', function(adminConfig) {
                return adminConfig.entities.length > 0;
            }],
            collections: ['adminConfig', function(adminConfig) {
                return adminConfig.dashboard().collections();
            }],
            // etc
```

It's convoluted, but it works. 

But: we're still injecting `NgAdminConfiguration` to a few places where I think it'll be hard to take them out:

```
> git grep NgAdminConfiguration
src/javascripts/ng-admin/Crud/column/maColumn.js:60:maColumn.$inject = ['$state', '$compile', 'NgAdminConfiguration', 'FieldViewConfiguration'];
src/javascripts/ng-admin/Crud/column/maReferenceManyLinkColumn.js:32:    maReferenceManyLinkColumn.$inject = ['$state', 'NgAdminConfiguration'];
src/javascripts/ng-admin/Crud/form/FormController.js:129:FormController.$inject = ['$scope', '$state', 'WriteQueries', 'NgAdminConfiguration', 'progression', 'notification', 'view', 'dataStore'];
src/javascripts/ng-admin/Crud/repository/ReadQueries.js:16:ReadQueriesService.$inject = ['RestWrapper', 'PromisesResolver', 'NgAdminConfiguration'];
src/javascripts/ng-admin/Crud/repository/WriteQueries.js:16:WriteQueriesService.$inject = ['RestWrapper', 'PromisesResolver', 'NgAdminConfiguration'];
```

examples/blog/config.js
```diff
@@ -1,438 +1,434 @@
...
+    });
+}]);
 
-        comment.editionView()
-            .fields(comment.creationView().fields())
-            .fields([nga.field(null, 'template')
-                .label('')
-                .template('<post-link entry="entry"></post-link>') // template() can take a function or a string
-            ]);
+var Factory = require('admin-config/lib/Factory');
```
fzaninotto: We can't make webpack a requirement at that stage. It's nice to support it, but the "Getting started" tutorial must be as simple as possible. Therefore, the definition of the DIC value must be in a `config()` or a `run()`, and get the factory from DIC.

examples/blog/config.js
```diff
@@ -1,438 +1,434 @@
...
+            factory.field('published', 'boolean').label('Is published ?')
+        ])
+        .listActions(['show'])
+        .order(4)
+    )
+    .template(customDashboardTemplate)
+);
+
+// add the admin configuration to angular's DI container, to make it availble from the view
+app.value('NgAdminConfiguration', admin);
```
fzaninotto: In order to support a fallback for the previous method (where the user doesn't define this value), you must name it differently, and keep the same meaning as before for `NgAdminConfiguration`. So the user can name the value the way she wants, then use the same name in the `<ma-ng-agmin>` configuration parameter. If there is no such parameter, the fallback is to use the global `NgAdminConfiguraiton` service, which must be defined by default.

examples/blog/config.js
jpetitcolas: Sure, that's just a PoC for the moment. Will clean it once everything work.

fzaninotto:
Since it really makes configuration harder for the devs, we'll have to support a BC way through `NgAdminConfiguration`. Maybe we'll revisit this one later, but for the time being, the idea is abandoned.

Issue #525, opened 2015-06-23
Access sub-attribute via dot notation ("address.street")


Xennis:
Hey,

my API returns data in the format (extract only):

``` js
[
        {
            "id": 1,
            "address": {
                "street": "Heidberg",
                 "postal_code": "75385"
            }
        },
        {
            "id": 2,
            "address": {
                "street": "Heinrich-Böll-Str.",
                "postal_code": "70707"
            }
        },
        ...
]
```

How can I access for instance the `street` attribute? It would be nice, if I could access it directly by `nga.field('address.street')`, but that does not seems to work. Of course I could define the field as `template` field and write a directive. But when I've many of such fields, that would not be so nice. Or is there another, simple way?

karantan:
I think this is what you are looking for:

```
app.config(function(RestangularProvider) {
    RestangularProvider.addElementTransformer('books', function(element) {
        for (var key in element.values) {
            element[key] = element.values[key];
        }

        return element;
    });
});
```
I copy paste this from the [documentation](https://github.com/marmelab/ng-admin/blob/master/doc/API-mapping.md)

jpetitcolas:
Indeed, we can do it currently the way @karantan suggested. ng-admin doesn't support the dot notation. But that's indeed a good idea. I keep the issue opened to not forget it. :)

fzaninotto:
Related to #68 and #390

fzaninotto:
Good news: that's now possible in the master branch, thanks to #549 being merged.

PR #549, opened 2015-07-07
[RFR] Clarify Map logic, allow nested fields

fzaninotto:
Currently, callbacks registered on fields with `map()` are applied to the REST response to turn it into an Entry (when fetching results) AND to the entries to turn them into a response (when pushing modifications). It doesn't make sense, and prevents some new features, such as deeply nested fields (refs. #525).

This PR fixes that. Now you can do things like:

```js
listView.fields([
    nga.field('title'),
    nga.field('author.name')
]);
```

to map REST responses looking like:

```json
[
  {
    "title": "War and Peace",
    "author": {
      "name": "Leo Tolstoi"
    }
  },
  {
    "title": "Anna Karenina",
    "author": {
      "name": "Leo Tolstoi"
    }
  }
]
```

Depends on https://github.com/marmelab/admin-config/pull/20

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "grunt-contrib-uglify": "~0.6.0",
     "grunt-contrib-watch": "~0.5.2",
     "grunt-exec": "^0.4.6",
-    "grunt-json-server": "git://github.com/fzaninotto/grunt-json-server.git#c834e50e2151a16930f1378324cf96b91438bdae",
+    "grunt-json-server": "git://github.com/fzaninotto/grunt-json-server.git",
     "grunt-karma": "^0.8.3",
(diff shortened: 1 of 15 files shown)


fzaninotto:
Switching to RFR!

package.json
```diff
@@ -13,7 +13,7 @@
     "npm": "^2.10.0"
   },
   "devDependencies": {
-    "admin-config": "^0.2.4",
+    "admin-config": "marmelab/admin-config#map_transform",
```
jeromemacias: `admin-config` `0.2.5` version had been published.

fzaninotto: Switched dependency to fixed version

PR #555, opened 2015-07-13
Ability to use custom method for get requests

lukemarsh:
I have just submitted a pull request for your admin-config module for adding a custom method to get requests. I have changed getOne and getList functions to allow for the custom methods

### DIFF
--- a/src/javascripts/ng-admin/Crud/misc/RestWrapper.js
+++ b/src/javascripts/ng-admin/Crud/misc/RestWrapper.js
@@ -16,13 +16,13 @@ define(function () {
      *
      * @returns {promise}
      */
-    RestWrapper.prototype.getOne = function(entityName, url) {
-        return this.Restangular
-            .oneUrl(entityName, url)
-            .get()
-            .then(function (response) {
-                return response.data;
-            });
+    RestWrapper.prototype.getOne = function(entityName, url, method) {
+        var resource = this.Restangular.oneUrl(entityName, url),
+            operation = method ? resource.customOperation(method, null, {}, {}, null) : resource.customGET();
+
+        return operation.then(function (response) {
+            return response.data;
+        });
     };
 
     /**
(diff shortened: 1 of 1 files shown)


fzaninotto:
Could you document the change in the docs? (cf 54f50e8e9b647966014dec3ec91bc6282cc57b65)

fzaninotto:
Refs https://github.com/marmelab/admin-config/pull/24

jpetitcolas:
Good to me. Waiting for [marmelab/admin-config#24](https://github.com/marmelab/admin-config/pull/24) PR to be merged.

jpetitcolas:
Can you just rebase on master to ensure all tests are green? We fixed them yesterday. :)

fzaninotto:
Your PR absolutely doesn't work. Did you even test it? The restangular code doesn't make sense.

Please, test your code before submitting it (e.g. by setting the retrieveMethod to 'GET' to make sure it doesn't break anything), I just spent half an hour trying to make this work.

lukemarsh:
Apologies for wasting your time on this one! I have since fixed said issues but the tests were failing due to some dependency issues, which I have fixed with this PR: https://github.com/marmelab/ng-admin/pull/585

fzaninotto:
Did you test with a simple `posts.retrieveMethod('GET')` in the example to check that it works?

lukemarsh:
Yes I did and it works :) but the method needs to be lowercase!

CKUDG:
Is this now fully implemented and works now?

I would also really need that feature to force POST method on the getList()...

Thanks much much and best regards;
C

mosasiru:
> Did you test with a simple posts.retrieveMethod('GET') in the example to check that it works?

It seems that `posts.retrieveMethod('get')` changes`getList` operation to 'get', so it does not work.
This test is edge case, because many uses case is that both getList and get operation are wanted to be  POST method, not GET method.
If you want it,  I will write pull-request that devides `retrieveMethod` into `getMethod` and `getListMethod`, but It would be too much. Besides, the name of `getListMethod` seems strange because it's operation.
or I can just fix conflict of it. Which do you prefer?

Issue #540, opened 2015-06-30
Error: $interpolate:interr


karantan:
When I .field() in createView() or editView() I get this error:

```
Error: [$interpolate:interr] Can't interpolate: row-{{ field.name() }}
TypeError: v2.name is not a function
http://errors.angularjs.org/1.4.1/$interpolate/interr?p0=row-%7B%7B%20field.name()%20%7D%7D&p1=TypeError%3A%20v2.name%20is%20not%20a%20function
    at angular.js:68
    at Function.$interpolateMinErr.interr (angular.js:10579)
    at interpolationFn (angular.js:10854)
    at attrInterpolatePreLinkFn (angular.js:8525)
    at invokeLinkFn (angular.js:8642)
    at nodeLinkFn (angular.js:8121)
    at compositeLinkFn (angular.js:7534)
    at nodeLinkFn (angular.js:8137)
    at compositeLinkFn (angular.js:7534)
    at compositeLinkFn (angular.js:7538)
```
 and this error:

```
Error: [$interpolate:interr] Can't interpolate: {{ field.name() }}
TypeError: v2.name is not a function
http://errors.angularjs.org/1.4.1/$interpolate/interr?p0=%7B%7B%20field.name()%20%7D%7D&p1=TypeError%3A%20v2.name%20is%20not%20a%20function
    at angular.js:68
    at Function.$interpolateMinErr.interr (angular.js:10579)
    at interpolationFn (angular.js:10854)
    at attrInterpolatePreLinkFn (angular.js:8525)
    at invokeLinkFn (angular.js:8642)
    at nodeLinkFn (angular.js:8121)
    at compositeLinkFn (angular.js:7534)
    at nodeLinkFn (angular.js:8137)
    at compositeLinkFn (angular.js:7534)
    at nodeLinkFn (angular.js:8137)
```

for every field I add. However this doesn't happen when I use my own template fields (via my directives).

Everything works fine and it doesn't break any code (as far as I know) but it still bothers me...
![screenshot 2015-06-30 17 07 01](https://cloud.githubusercontent.com/assets/7062631/8434314/fdb9d892-1f4a-11e5-83ef-1d15ffa41350.png)

fzaninotto:
Could you share your config?

karantan:
https://www.dropbox.com/s/wgbeu4rt9qvt2yf/admin.js?dl=0

jpetitcolas:
[...]

PR #563, opened 2015-07-21
Error: $interpolate:interr #540

FairlieAgile:
Replacing name() with label() resolves the issue


jpetitcolas:
Thanks for this PR, but I don't think this is the correct way to fix it. For instance, if I have a field named `last_update` and I want to label it as `Last update date`, then I would have strange id: `row-Last update date`.

jpetitcolas:
It is an Angular 1.4 issue. Some clues about this error: http://stackoverflow.com/questions/24492062/how-to-resolve-all-the-interpolation-errors

Closing this PR as it is not the good way to fix it, as explained above. :)

PR #564, opened 2015-07-22
[makefile] Allow building devel files

manuelnaranjo:
Allow building devel non minified versions of the library so debugging is easier for non core developers, IMO this files should be part of the upstream package. Sometimes it helps to see what's going on underneath the courtains.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -8,6 +8,14 @@ function getEntrySources(sources) {
     return sources;
 }
 
+function getOutputName(extension) {
+    if (process.env.NODE_ENV === 'development') {
+        return 'build/[name].' + extension;
+    }
+
+    return 'build/[name].min.' + extension;
+}
+
 var ngAdminSources = [
     './src/javascripts/ng-admin.js',
     './src/sass/ng-admin.scss'
@@ -32,7 +40,7 @@ module.exports = {
     },
     output: {
         publicPath: "http://localhost:8080/",
-        filename: "build/[name].min.js"
+        filename: getOutputName('js')
     },
     module: {
         loaders: [
@@ -45,7 +53,7 @@ module.exports = {
         ]
     },
     plugins: [
-        new ExtractTextPlugin('build/[name].min.css', {
+        new ExtractTextPlugin(getOutputName('css'), {
             allChunks: true
         })
     ]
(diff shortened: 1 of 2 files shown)


jpetitcolas:
Thanks for this PR. However, I think we should embed non minified sources with every build. This way, developers would be able to debug more easily their application without needing to clone this project.

As a consequence, can you update the `build` task to make a `webpack -p` AND a `webpack` of the sources?

manuelnaranjo:
@jpetitcolas it's done now `make build` depens on devel getting both versions at the same time

webpack.config.js
```diff
@@ -1,13 +1,22 @@
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 function getEntrySources(sources) {
-    if (process.env.NODE_ENV !== 'production') { // for live reload
+    if (process.env.NODE_ENV !== 'production' &&
+        process.env.NODE_ENV !== 'development') { // for live reload
```
jpetitcolas: We need the live reload in development.

manuelnaranjo:
Ok then, I will make a different env called non-minified what do you think? so it doesn't collide with production or development

jpetitcolas:
IMO, a better solution would be:

* Change the `build` command to do both `webpack` and `webpack -p` commands
* Update the `webpack.config.js` file to update filenames: `ng-admin.min.js` if using the production flag, `ng-admin.js` otherwise

To detect the `-p` flag, you can use `process.argv`, as pinpointed by @bebraw on the [Webpack Gitter](https://gitter.im/webpack/webpack). :)

This way, you wouldn't need any extra environment.

jpetitcolas:
Looks good to me. Can you rebase, to ensure tests are passing?

By the way, do not hesitate to ping us when your PR has been updated. I wasn't aware you made some changes. :)

jpetitcolas:
Fixes #614

fzaninotto:
I've tested this PR, and the non-minified files are not helping, because they are not readable. Since we use babel for transpiling ES6 scripts, the non-minified code is already hard to read.

The good solution to your problem, I think, are sourcemaps.

manuelnaranjo:
sourcemaps are way better, I've been using a wiredep solution now, I will create a reference proyect soon and share.

Yet sourcemap are not been provided, not even generated by the current setup.

fzaninotto:
I'll work on a sourcemap setup.

fzaninotto:
Superseded by #619

PR #568, opened 2015-07-23
[RFR] Fix reference_many field when value is []

### LINKED ISSUE #52: Field type file?
Hi! This project will be very useful for me, really thank you!

I Would like to know if is in your to-do list allow user to upload some kind of file in server or s3 etc etc...

I think this could be very useful :)

Thanks

### LINKED ISSUE #552: bug: reference_many field
I found a bug with reference_many field. When you edit e.g. Post and remove all Tags then save this Post you cannot assign any Tags to this Post because they do not appear.

Here is a link to the video that shows this bug: http://www.youtube.com/watch?v=iohsdt2VBG0

edit: this bug appreas only in version 0.8.0. In 0.7.0 this is OK because this field is a list...

jpetitcolas:
Fixes #552

### DIFF
--- a/examples/blog/config.js
+++ b/examples/blog/config.js
@@ -117,12 +117,7 @@
                     .targetEntity(tag)
                     .targetField(nga.field('name'))
                     .filters(function(search) {
-                        if (!search) {
-                            return;
-                        }
-                        return {
-                            q: search
-                        }
+                        return search ? { q: search } : null;
                     })
                     .remoteComplete(true, { refreshDelay: 300 })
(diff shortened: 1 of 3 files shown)


examples/blog/config.js
```diff
@@ -117,6 +117,9 @@
                     .targetEntity(tag)
                     .targetField(nga.field('name'))
                     .filters(function(search) {
+                        if (!search) {
```
fzaninotto: ```return search ? { q: search } : null;```

jpetitcolas:
Code review applied. Ready for merge.

PR #573, opened 2015-07-25
Fix date field write widget

manuelnaranjo:
Date field write widget was inside a wrong container, making it display ugly

### DIFF
--- a/src/javascripts/ng-admin/Crud/fieldView/DateFieldView.js
+++ b/src/javascripts/ng-admin/Crud/fieldView/DateFieldView.js
@@ -11,7 +11,7 @@ define(function(require) {
         return '<ma-date-field field="::field" value="values[field.name()]"></ma-date-field>';
     }
     function getWriteWidget() {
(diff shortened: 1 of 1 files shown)


fzaninotto:
What? No, limiting the date widget width was done on purpose, and it works fine. See the example from the demo:

image

Can you show us the visual difference between before and after your PR?

manuelnaranjo:
I had an example, if I reproduce it again I post it here, want to close this until then?

I've been thinking that row, label and field classes should be configurable, I've been thinking on how to implement this, when I make a demo I open an issue.

manuelnaranjo:
Found it!

ng-admin-dates

I can give you guys access to our server so you can look at it by your selfs if anyone is willing to look at it. Just look for me at gitter.

fzaninotto:
Did you customize the CSS?

manuelnaranjo:
@fzaninotto nope, the date field is set to col-sm-4 by default which is too short for displaying this dates.

fzaninotto:
Superseded by #673

PR #574, opened 2015-07-25
Imprve Reference Field View Widget

manuelnaranjo:
Reference field should use ma-column instead of ma-string-column so the
    formatting requested by the programmer is honored.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "npm": "^2.10.0"
   },
   "devDependencies": {
-    "admin-config": "^0.2.5",
+    "admin-config": "manuelnaranjo/admin-config#reference-field-target-entries",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "npm": "^2.10.0"
   },
   "devDependencies": {
-    "admin-config": "^0.2.5",
+    "admin-config": "manuelnaranjo/admin-config#reference-field-target-entries",
```
fzaninotto: You'll want to PR admin-config first, we can't merge that.

manuelnaranjo:
@fzaninotto ok will do

fzaninotto:
Superseded by #688

PR #621, opened 2015-08-25
[RFR] Fix file field

### LINKED ISSUE #597: File field for file upload
I am using ng-admin file field for file upload. 
When I use ng-admin-master version 24/06/2015 everything is working ok. In all browsers: IE, FF, Chrome.
But when I try to use later ng-admin-master version (for example 19/07/2015 or 06/08/2015) file upload does not work. In all browsers also.
What is wrong? What can cause this problem?

jeromemacias:
Fix #597

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maFileField.js
+++ b/src/javascripts/ng-admin/Crud/field/maFileField.js
@@ -52,7 +52,6 @@ define(function (require) {
                     }
 
                     scope.fileSelected = function(selectedFiles) {
-                        console.log(selectedFiles);
                         if (!selectedFiles || !selectedFiles.length) {
                             return;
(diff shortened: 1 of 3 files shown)


src/javascripts/ng-admin/Crud/field/maFileField.js
```diff
@@ -52,6 +52,7 @@ define(function (require) {
                     }
 
                     scope.fileSelected = function(selectedFiles) {
+                        console.log(selectedFiles);
```
fzaninotto: debug spotted

jeromemacias:
@fzaninotto Tests are ok, fail is due to data that had been changed on the public demo...

Issue #320, opened 2015-02-19
Mock backend to let e2e tests run without backend


fzaninotto:
PRs started on a fork have failed e2e tests. We definitely need to update e2e tests using a `$httpBackend` mock to avoid depending on a third-party backend.

fzaninotto:
To be done using https://github.com/marmelab/FakeRest

fzaninotto:
Now done!

PR #625, opened 2015-08-27
[RFR] Use fakerest and webpack-dev-server instead of json-server and grunt-connect

jeromemacias:
* [x] Use Fakerest and webpack-dev-server for demo
* [x] Use Fakerest for e2e tests
* [x] Fix unit tests (using sinon 1.14.1 instead of sinon-2.0)
* [x] Use latest `fakerest`
* [x] Clean `Gruntfile.js`


Makefile
```diff
@@ -5,7 +5,7 @@ install:
 	./node_modules/protractor/bin/webdriver-manager update
 
 run:
-	@grunt
+	@./node_modules/webpack-dev-server/bin/webpack-dev-server.js --colors --devtool cheap-module-inline-source-map --content-base examples/blog --port 8000
```
fzaninotto: don't forget to remove the related conf in Gruntfile

fzaninotto:
Awesome!

Makefile
jeromemacias: Done

examples/blog/build/fakerest.js
```diff
@@ -0,0 +1 @@
+../../../node_modules/fakerest/dist/FakeRest.min.js
```
fzaninotto: I don't understand what that file does?

examples/blog/fakerest-init.js
```diff
@@ -0,0 +1,31 @@
+(function () {
+    'use strict';
+
+    // setup fake server
+    var restServer = new FakeRest.Server('http://localhost:3000');
+    var testEnv = window.location.pathname.indexOf('test.html') !== -1;
+    restServer.init(apiData);
```
fzaninotto: apiData is global?

fzaninotto: Nevermind, I saw what you loaded the data in a `<script>` tag before fakerest in index.html.

fzaninotto:
Let me add another item:

-[ ] Add a task to compile and save the JS so that the entire demo can be run from a static server (think gh-pages)

examples/blog/build/fakerest.js
jeromemacias: It's just a symlink to the fakerest node module, to be able to import it in the `<script>` of the demo index.

fzaninotto: Ah OK, that's how GitHub show symlinks, I didn't know

Gruntfile.js
```diff
@@ -15,9 +15,17 @@ module.exports = function (grunt) {
                 dest: 'src/javascripts/test/fixtures/',
                 options: {
                     process: function(content) {
-                        return content.replace(/http\:\/\/localhost\:8080\//g, '/');
+                        return content.replace(/http\:\/\/localhost\:8000\//g, '/');
                     }
                 }
+            },
+            test_fakerest: {
+                src: 'node_modules/fakerest/dist/FakeRest.min.js',
```
fzaninotto: why do you need to copy if there is a symlink?

jeromemacias: Because copy from example build directory to test build directory does not copy symlinks (which would be broken anyway).

fzaninotto:
Closes #320

fzaninotto:
Could you rebase?

jeromemacias:
Rebased

src/javascripts/test/e2e/filterViewSpec.js
```diff
@@ -165,7 +165,7 @@ describe('List filter', function () {
             $$('.filters .filter:nth-child(1) input').sendKeys('be');
             browser.driver.sleep(600); // debounce delay
             $$('ma-datagrid-pagination .total').then(function (totalElements) {
-                expect(totalElements[0].getText()).toBe('1 - 5 on 5');
+                expect(totalElements[0].getText()).toBe('1 - 4 on 4');
```
fzaninotto: Why does it change if the dataset is the same? Something's wrong.

jeromemacias: I don't know how it's possible to find 5 records, see http://ng-admin.marmelab.com/#/comments/list?search={%22q%22:%22be%22}&page=1

But if I search for `b` instead of `be`, 5 record are found, maybe a coincidence, but this is the only explaination I see, protractor wasn't wait for the `e` with the old basecode.

fzaninotto: OK, FakeRest doesn't handle the q search inside composite fields (the 5th comment has an author.name containing 'be'.

PR #639, opened 2015-09-01
[RFR] Introducing filter default value

fzaninotto:
You can now use `defaultValue()` on filter fields, so as to filter the list as soon as the filter is added. Combined with an empty template, this allows to create "tagged" lists:

```js
var user_id = 123; // currently logged user
var d = new Date()
var yesterday = d.setDate(d.getDate() - 1);
listView.filters([
    nga.field('flagged', 'template')
        .defaultValue('true'), // adds ?flagged=true to the REST query
    nga.field('author_id', 'template')
        .label('Mine')
        .defaultValue(user_id), // adds ?author_id=123 to the REST query
    nga.field('created_at', 'template')
        .label('Recent')
        .defaultValue({ gt: yesterday }), // adds ?created_at={gt:2015-08-31} to the REST query
]);
```


src/javascripts/ng-admin/Crud/filter/maFilter.js
```diff
@@ -33,10 +32,15 @@ function maFilterDirective(FieldViewConfiguration) {
         scope: {
             filters: '=',
             datastore: '&',
-            values: '&'
+            values: '&',
```
fzaninotto: There is a mystery that I can't explain: `values` uses one-way databinding in an isolated scope, yet the watcher on `ListLayoutController` gets notified when the `values` change. Weird, but works.

PR #642, opened 2015-09-01
[RFR] Return to list view when deleting a record

### LINKED ISSUE #641: 404 after successful delete
If Delete button is pressed on the details page (ShowView) and delete is successful the user will be taken back to the details page which no longer exists, resulting in 404. What should happen is that the entity listing is displayed.

Try it by deleting from the following url http://ng-admin.marmelab.com/#/posts/show/1

jpetitcolas:
Needs to implement and fix tests.

Closes #641.

### DIFF
--- a/src/javascripts/ng-admin/Crud/delete/DeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/DeleteController.js
@@ -3,9 +3,10 @@
 define(function () {
     'use strict';
 
-    var DeleteController = function ($scope, $window, WriteQueries, notification, params, view, entry) {
+    var DeleteController = function ($scope, $window, $state, WriteQueries, notification, params, view, entry) {
         this.$scope = $scope;
         this.$window = $window;
+        this.$state = $state;
         this.WriteQueries = WriteQueries;
         this.entityLabel = params.entity;
         this.entityId = params.id;
@@ -22,21 +23,29 @@ define(function () {
 
     DeleteController.prototype.deleteOne = function () {
         var notification = this.notification,
-            entityName = this.entity.name(),
-            $window = this.$window;
+            entityName = this.entity.name();
 
-        this.WriteQueries.deleteOne(this.view, this.entityId).then(function () {
-            this.back();
-            notification.log('Element successfully deleted.', { addnCls: 'humane-flatty-success' });
-        }.bind(this), function (response) {
-            // @TODO: share this method when splitting controllers
-            var body = response.data;
-            if (typeof body === 'object') {
-                body = JSON.stringify(body);
-            }
+        return this.WriteQueries.deleteOne(this.view, this.entityId)
+            .then(
+                function () {
+                    this.$state.go(this.$state.get('list'), angular.extend({
+                        entity: entityName
+                    }, this.$state.params));
 
-            notification.log('Oops, an error occured : (code: ' + response.status + ') ' + body, {addnCls: 'humane-flatty-error'});
-        });
+                    notification.log('Element successfully deleted.', { addnCls: 'humane-flatty-success' });
+                }.bind(this),
+                function (response) {
+                    // @TODO: share this method when splitting controllers
+                    var body = response.data;
+                    if (typeof body === 'object') {
+                        body = JSON.stringify(body);
+                    }
+
+                    this.back();
+
+                    notification.log('Oops, an error occured : (code: ' + response.status + ') ' + body, {addnCls: 'humane-flatty-error'});
+                }
+            );
     };
 
     DeleteController.prototype.back = function () {
--- a/src/javascripts/ng-admin/Crud/delete/DeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/DeleteController.js
@@ -3,7 +3,7 @@
 define(function () {
     'use strict';
 
-    var DeleteController = function ($scope, $window, $state, WriteQueries, notification, params, view, entry) {
+    var DeleteController = function ($scope, $window, $state, $q, WriteQueries, notification, params, view, entry) {
         this.$scope = $scope;
         this.$window = $window;
         this.$state = $state;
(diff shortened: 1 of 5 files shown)


src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -22,21 +23,29 @@ define(function () {
...
+                    notification.log('Element successfully deleted.', { addnCls: 'humane-flatty-success' });
+                }.bind(this),
+                function (response) {
+                    // @TODO: share this method when splitting controllers
+                    var body = response.data;
+                    if (typeof body === 'object') {
+                        body = JSON.stringify(body);
+                    }
+
+                    this.back();
```
fzaninotto: Why do you go back in case of error? Doesn't make sense.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -22,21 +23,29 @@ define(function () {
...
-            notification.log('Oops, an error occured : (code: ' + response.status + ') ' + body, {addnCls: 'humane-flatty-error'});
-        });
+            entityName = this.entity.name();
+
+        return this.WriteQueries.deleteOne(this.view, this.entityId)
+            .then(
+                function () {
+                    this.$state.go(this.$state.get('list'), angular.extend({
+                        entity: entityName
+                    }, this.$state.params));
```
fzaninotto: `this.$state.params` are empty in the `DeletionView`. Why do you need it?

fzaninotto:
One thing that concerns me is that if you're coming from a filtered list you'll lose the filter. 

I think you should be smarter and keep the `back()` when coming from a list, and going to the list otherwise.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
jpetitcolas: Because of https://github.com/marmelab/ng-admin/pull/538

But true, there will be another issue here: what about deleting a record from a reference_many list?

A solution may be to add a "with_redirect" to true by default, and open the confirmation box in a pop-up. Reference many delete button would not redirect in this case. What do you think of it?

jpetitcolas:
After an IRL discussion, the most easy way to achieve it is to always redirect to the previous page.

We only redirect to the list view if previous page has the same entity and id than the deleted object. In this case, previous view doesn't exist anymore, whatever it is (show or edition view).

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -17,26 +18,48 @@ define(function () {
         this.notification = notification;
         this.$scope.entry = entry;
         this.$scope.view = view;
+
         $scope.$on('$destroy', this.destroy.bind(this));
+
+        this.previousStateParametersDeferred = $q.defer();
+        $scope.$on('$stateChangeSuccess', (event, to, toParams, from, fromParams) => {
+            console.log(fromParams);
```
jpetitcolas: Console log spotted

src/javascripts/ng-admin/Crud/delete/DeleteController.js
jpetitcolas: To remove

jpetitcolas:
Switching to RFR.

jpetitcolas:
Tests should now be green: I removed `sinon` dependency which caused some troubles with Webpack (it was not used before this PR).

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -17,26 +18,45 @@ define(function () {
         this.notification = notification;
         this.$scope.entry = entry;
         this.$scope.view = view;
+
         $scope.$on('$destroy', this.destroy.bind(this));
+
+        this.previousStateParametersDeferred = $q.defer();
+        $scope.$on('$stateChangeSuccess', (event, to, toParams, from, fromParams) => {
+            this.previousStateParametersDeferred.resolve(fromParams);
```
fzaninotto: How about if I click a link in the navbar rather than deleting?

fzaninotto: You need to bind once, or stop the listener after deletion.

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -17,26 +18,45 @@ define(function () {
...
+
+                        // if previous page was related to deleted entity, redirect to list
+                        if (previousStateParameters.entity === entityName && previousStateParameters.id === this.entityId) {
+                            return this.$state.go(this.$state.get('list'), angular.extend({
+                                entity: entityName
+                            }, this.$state.params));
+                        }
+
+                        this.back();
+                    }.bind(this));
```
fzaninotto: You don't need that if you use `() => {}`

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -17,26 +18,45 @@ define(function () {
...
-
-            notification.log('Oops, an error occured : (code: ' + response.status + ') ' + body, {addnCls: 'humane-flatty-error'});
-        });
+            entityName = this.entity.name();
+
+        return this.WriteQueries.deleteOne(this.view, this.entityId)
+            .then(
+                function () {
+                    this.previousStateParametersDeferred.promise.then(function(previousStateParameters) {
+                        notification.log('Element successfully deleted.', { addnCls: 'humane-flatty-success' });
```
fzaninotto: Show the notification after changing the state

fzaninotto:
unit tests fail

src/javascripts/ng-admin/Crud/delete/DeleteController.js
jpetitcolas: It would be unbound at controller destruction (as bound with `$on`).

jpetitcolas:
Green! \o/

fzaninotto:
Thanks!

PR #644, opened 2015-09-01
[RFR] Allow the main menu to use a custom template

fzaninotto:


cmnstmntmn:
@fzaninotto i am trying to remap the content of **menu.js** to a custom template.
 
1. i changed the whole **layout.html**

````
    admin.layout(require('./layout.html'));
    admin.header(require('./header.html'));
    admin.menu(require('./menu')(nga, admin));
    admin.dashboard(require('./dashboard/config')(nga, admin));
````

2. somewhere in **layount.html**, i am trying to 

 ``<span compile="::appController.menu"></span>``

i only get an array, becauze the template, is missing from the equation.
screen shot 2015-09-14 at 22 14 16


---

i get no luck even if put hte content of [menuBar.html](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Main/view/menuBar.html) inside **layout.html**

my question is: can this **admin.menu()** accep both data, and a custom template ?

thank you

fzaninotto:
`appController.menu` is [a `Menu` object](https://github.com/marmelab/admin-config/blob/b4b057490b7bb35928fb08c9d7160751552ba4bf/lib/Menu/Menu.js). Don't try to output it in a template, it doesn't make any sense. It contains both menus (data) and the template.

Why don't you use the `<ma-menu-bar>` directive?

cmnstmntmn:
well, it's more a markup/css issue ..

having **ul.nav** nested in two other divs, makes things a bit hard on css side.

screen shot 2015-09-17 at 21 14 30

i wanted to achieve this kind of markup for the aside bar

aside

but having both **header** and `<ma-menu-bar>`  inside ``<nav`` makes it tricky

fzaninotto:
Then do the markup and the directives you want. But don't try to output a JS object to HTML, it doesn't make sense.

cmnstmntmn:
does ``<ma-menu-bar>`` can be extended?
i think the easiest way to achive this, is to append a custom mark-up to ``<ma-menu-bar>`` directive

fzaninotto:
You can customize the template of every directive in ng-admin (and in Angular.js in general). 

https://github.com/marmelab/ng-admin/blob/master/doc/Theming.md#customizing-directives-templates

PR #651, opened 2015-09-02
[RFR] Fix bug where filters are freezed when url changes

fzaninotto:


jpetitcolas:
Tests fails in List Filters: 

>  interaction with pagination
    should reset page number - fail

PR #652, opened 2015-09-02
ReferenceRefresher : Add check on field shouldRemoveDuplicates method

### LINKED ISSUE #649: Extending ReferenceManyField with new type cause dupes choices to appear
Hi!
I managed to extend `ReferenceManyField` like this : 
```
import ReferenceManyField from 'admin-config/lib/Field/ReferenceManyField';
class ReferenceManyCustomField extends ReferenceManyField {
    constructor(name) {
        super(name);
        this._type = 'reference_many_custom';
    }
}
export default ReferenceManyCustomField;
```
But now the `choices` displayed include the current field value.
If I remove the line where I change the `_type` property, it goes back to normal.
Any hints on why this is happening?

Benew:
Fixes #649

### DIFF
--- a/src/javascripts/ng-admin/Crud/repository/ReferenceRefresher.js
+++ b/src/javascripts/ng-admin/Crud/repository/ReferenceRefresher.js
@@ -11,7 +11,7 @@ class ReferenceRefresher {
             .then(r => r[field.name()])
             .then(results => this._transformRecords(field, results));
 
-        if (field.type() === 'reference_many' || field.type() === 'choices') {
+        if (field.type() === 'reference_many' || field.type() === 'choices' || ( field.shouldRemoveDuplicates && field.shouldRemoveDuplicates() ) ) {
             promise = promise.then(formattedResults => this._removeDuplicates(formattedResults, currentValue));
         }
 


Benew:
The build fails... It's just because I didn't commit the built files right?

fzaninotto:
Nope, Travis rebuilds everything from scratch, you never have to commit build files.

src/javascripts/ng-admin/Crud/repository/ReferenceRefresher.js
```diff
@@ -11,7 +11,7 @@ class ReferenceRefresher {
             .then(r => r[field.name()])
             .then(results => this._transformRecords(field, results));
 
-        if (field.type() === 'reference_many' || field.type() === 'choices') {
+        if (field.type() === 'reference_many' || field.type() === 'choices' || ( field.shouldRemoveDuplicates && field.shouldRemoveDuplicates() ) ) {
```
fzaninotto: This method doesn't exist in current admin-config. If you're depending on a change from another library, you must update the `package.json` to mention the dependency.

PR #653, opened 2015-09-02
[RFR] Expose field constructors to avoid Webpack/Babel fuss

fzaninotto:
It becomes possible to recuperate a field constructor without installing the `admin-config` module:

```
var NumberField = nga.getFieldConstructor('number');

// now create a new class extending NumberField
function AmountField(name) {
  NumberField.call(this, name)
}
AmountField.prototype = new NumberField();
AmountField.prototype.constructor = AmountField;
// or something like that
```


Benew:
This is pretty cool! Thanks!

manuelnaranjo:
pretty cool feature!

Benew:
Thanks guys! Your test/build workflow is really worth the fuss though!
Learned a lot from it! I moved to "vanilla" angular material because of the
apparent lack of Rest provider access since then. There is something
exciting there too!
Cheers!

PR #667, opened 2015-09-07
[RFR] Menu autoclose

ThieryMichel:


jpetitcolas:
Can you add some tests?

doc/Menus.md
```diff
@@ -46,12 +46,16 @@ admin.menu(nga.menu()
 );
 ```
 
-*Tip*: `admin.menu()` is both a setter and a getter. You can modify an existing menu in the admin configuration by using `admin.menu().getChildByTitle()`
+By default parent menus will automatically close when none of their children are active. This can be deactivated with:
+```
+nga.menu().autoClose(false);
+```js
```
fzaninotto: bad formatting: the 'js' must be on the opening line. Also, please add a newline after the code snippet.

src/javascripts/ng-admin/Main/component/directive/maMenuBar.js
```diff
@@ -10,7 +10,10 @@ function maMenuBar($location, $rootScope, $compile) {
         link: function(scope, element) {
             scope.menu = scope.menu();
             scope.path = $location.path();
-            var openMenus = [];
+            // initialize openMenus
+            var openMenus = scope.menu.children().filter(function (menu) {
```
fzaninotto: I don't understand why you need that, it already works by way of `isActive()`

ThieryMichel: It is not properly initialized. If we refresh on a link that open a child menu, the child menu will not be in the openMenus. And so the menu will close when navigating outside of the submenu even when autoClose is false.

Issue #383, opened 2015-04-06
Display boolean field of referenced type in listView or dashboardView


ihrigb:
Is there any posibility to display a boolean field of a referenced type in the listView or dashboardView?

This is the approach I tried, but it didn't work out:

```
entity.listView()
    .fields([
        nga.field('id', 'number').label('ID'),
        nga.field('name', 'string'),
        nga.field('referenced_entity', 'reference')
            .targetEntity(referencedEntity)
            .targetField(nga.field('option', 'boolean'))
    ]);
```

fzaninotto:
No, it's currently not possible for now. A refactoring of the internals of references is necessary to enable it. It's on our todo list.

fzaninotto:
I don't know since when, but now it works fine.

PR #688, opened 2015-09-11
[RFR] Preload references in refrenced_list entries

### LINKED ISSUE #637: reference field in referenced_list?
Is it possible to have a reference field within a referenced list? eg:

```javascript
nga.field('accounts', 'referenced_list') // display list of related accounts
    .targetEntity(accounts) 
    .targetReferenceField('voip_id')
    .targetFields([
        nga.field('id').isDetailLink(true),
        nga.field('date_created'),
        nga.field('user_profile_id'), // <-- this works fine
        nga.field('user_profile_id', 'reference') // <-- it doesn't call the api request for the profile entity
            .label('User Profile')
            .validation({required: true})
            .targetEntity(profiles)
            .targetField(nga.field('username')),
    ])
```

fzaninotto:
- [x] Implement the feature
- [x] Refactor the routing for more readbility
- [x] make tests pass

Closes #637, #383

### DIFF
--- a/src/javascripts/ng-admin/Crud/column/maReferenceColumn.js
+++ b/src/javascripts/ng-admin/Crud/column/maReferenceColumn.js
@@ -1,0 +1,28 @@
+function maReferenceColumn() {
+    return {
+        restrict: 'E',
+        scope: {
+            field: '&',
+            value: '&',
+            datastore: '&'
+        },
+        link: {
+            pre: function(scope) {
+                const value = scope.value();
+                scope.field = scope.field();
+                scope.targetEntity = scope.field.targetEntity();
+                scope.targetField = scope.field.targetField();
+                const identifierName = scope.targetEntity.identifier().name()
+                scope.referencedEntry = scope.datastore()
+                    .getEntries(scope.targetEntity.uniqueId + '_values')
+                    .filter(entry => entry.values[identifierName] == value)
+                    .pop();
+            }
+        },
+        template: '<ma-column field="::targetField" entry="::referencedEntry" entity="::targetEntity" datastore="::datastore()"></ma-column>'
+    };
+}
+
+maReferenceColumn.$inject = [];
+
+export default maReferenceColumn;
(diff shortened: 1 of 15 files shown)


fzaninotto:
Switching to RFR

src/javascripts/ng-admin/Crud/column/maReferenceColumn.js
```diff
@@ -0,0 +1,28 @@
...
+        link: {
+            pre: function(scope) {
+                const value = scope.value();
+                scope.field = scope.field();
+                scope.targetEntity = scope.field.targetEntity();
+                scope.targetField = scope.field.targetField();
+                const identifierName = scope.targetEntity.identifier().name()
+                scope.referencedEntry = scope.datastore()
+                    .getEntries(scope.targetEntity.uniqueId + '_values')
+                    .filter(entry => entry.values[identifierName] == value)
```
jpetitcolas: `===`

src/javascripts/ng-admin/Crud/column/maReferencedListColumn.js
```diff
@@ -17,8 +17,9 @@ function maReferencedListColumn(NgAdminConfiguration) {
 <ma-datagrid name="{{ field.datagridName() }}"
     entries="::entries"
     fields="::field.targetFields()"
-    list-actions="::field.listActions()" 
-    entity="::entity">
+    list-actions="::field.listActions()"
+    entity="::entity"
+    datastore="::datastore()">
```
jpetitcolas: Can't we retrieve `entries` from the datastore now?

jpetitcolas: Got my response IRL: we can't remove entries, as this is the only way to keep entry ids.

jpetitcolas:
:+1:

src/javascripts/ng-admin/Crud/column/maReferenceColumn.js
fzaninotto: Assuming that the APIs we fetch are consistent (i.e. that the type of `comment.post_id` is the same as the type of `post.id`) is pretty brave, but I think we'd rather be tolerant to ease compatibility. So e.g. when `comment.post_id` is `'1'` and `post.id` is `1`, the current code works.

src/javascripts/ng-admin/Crud/column/maReferencedListColumn.js
fzaninotto: Yep, but we'd need to pass an array of ideas and a type to let the datagrid retrieve the entries from the datastore... not ideal.

Issue #635, opened 2015-08-31
datetime format is not being applied


jamespsterling:
Doing something like this with mysql datetime field, 

```javascript
    // submission
    submission.listView()
    .title('Submissions')
    .fields([
        nga.field('id').label('Id'),
        nga.field('dateSubmitted', 'datetime').format('MMM dd, yyyy').isDetailLink(true),
        nga.field('dateAccepted', 'datetime').format('MMM dd, yyyy'),
        nga.field('submissionMonth'),
        nga.field('submissionYear'),
```

The format clause is not being applied.

jamespsterling:
Example datetime, 2014-11-08 07:12:49

jpetitcolas:
This is a format issue. We apply the Angular `date` filter, which work only on some format:

> Date to format either as Date object, milliseconds (string or number) or various ISO 8601 datetime string formats (e.g. yyyy-MM-ddTHH:mm:ss.sssZ and its shorter versions like yyyy-MM-ddTHH:mmZ, yyyy-MM-dd or yyyyMMddTHHmmssZ). If no timezone is specified in the string input, the time is considered to be in the local timezone.

Try to update your date format adding a `T` between date and time, and a `Z` at the end. It should then work.

Documentation updated in #702.

jamespsterling:
To those interested and see this issue, here is an example of an Restangular response interceptor where the response translates the mysql datetime into angular accepted datetime.

```
    RestangularProvider.addResponseInterceptor(function(data, operation, what, url, response, deferred)
    {
        // response mapping
        if (what == "submissions")
        {
            angular.forEach(data, function(value, key) {

                value.dateSubmitted = Date.convertMysqlToAngular(value.dateSubmitted);
                value.dateUpdated = Date.convertMysqlToAngular(value.dateUpdated);
                value.dateAccepted = Date.convertMysqlToAngular(value.dateAccepted);
            });
        }
```

Here is the Date function,
```
Date.convertMysqlToAngular = function(mysql_string)
{ 
   var t, result = null;

   if( typeof mysql_string === 'string' )
   {
      t = mysql_string.split(/[- :]/);

      //when t[3], t[4] and t[5] are missing they defaults to zero
      result = new Date(t[0], t[1] - 1, t[2], t[3] || 0, t[4] || 0, t[5] || 0);          
   }

    if (result == null) {
        return;
    }

    // Grab each of your components
    var yyyy = result.getFullYear().toString();
[...]

PR #702, opened 2015-09-22
[RFR] Add doc on input date requirement for format

jpetitcolas:
Refs #635.


jpetitcolas:
Following a discussion with @jamespsterling on Gitter.

PR #704, opened 2015-09-23
Doc: fixed bad method name (Reference:targetLabel)

SebLours:
Reference:targetLabel -> Reference:targetField


fzaninotto:
Thanks!

PR #708, opened 2015-09-25
[RFR] Fix invalid reference to entry in datagrid header

djhi:
- [x] Don't specify `entry` or headers
- [x] Fix blog example to check for entry before using it in getCssClasses for tag published
- [x] Updated documentation

### DIFF
--- a/doc/Configuration-reference.md
+++ b/doc/Configuration-reference.md
@@ -455,11 +455,15 @@ Set field validation rules. Based on Angular's form validation features.
             } })
 
 * `cssClasses(String|Function)`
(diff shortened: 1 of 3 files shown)


fzaninotto:
I don't understand this PR. What is invalid?

djhi:
`ng-class="field.getCssClasses(entry)"` `entry` does not exist at that point

djhi:
And that was causing errors in the example app on the tags list view

fzaninotto:
Ah, you're right. But in that case, I suggest we keep the call, without passing `entry` as argument. This still allows to add a custom class to the header.

djhi:
Ok but shouldn't we document that case ?

fzaninotto:
Thanks!

Issue #624, opened 2015-08-26
infinitePagination does not resolve reference fields


toxsick:
Hey guys,

in my listView I use reference fields like this:

```
nga.field('make', 'reference')
    .map(truncate)
    .targetEntity(make)
    .targetField(nga.field('name').map(truncate))
```

This works absolutely fine with normal pagination. But if I use `.infinitePagination(true)` the references are only resolved for the first page. When the second page is loaded no calls are triggered to resolve the references. That's why only data is visible that had been already requested on the first page.

Here you can see what this looks like:

![angular_admin_und_datadock_und_datadock](https://cloud.githubusercontent.com/assets/951466/9500050/72d35d38-4c23-11e5-98af-58d6cdcd68b2.png)

HELP?

Best Regards
Hannes

fzaninotto:
That's a bug, we'll need to investigate. In the meantime, please use normal pagination.

PR #709, opened 2015-09-25
[RFR] infinitePagination now resolve reference fields

djhi:
Fix for #624

### DIFF
--- a/src/javascripts/ng-admin/Crud/list/ListController.js
+++ b/src/javascripts/ng-admin/Crud/list/ListController.js
@@ -1,99 +1,94 @@
-/*global define*/
 import Entry from 'admin-config/lib/Entry';
 
-define(function () {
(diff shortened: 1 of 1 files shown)


src/javascripts/ng-admin/Crud/list/ListController.js
```diff
@@ -1,4 +1,5 @@
 /*global define*/
+import Entry from 'admin-config/lib/Entry';
 
 define(function () {
```
jpetitcolas: Not specific to this PR, but can you remove the `define`, as it is just requirejs obsolete code? Just `export default` the controller function.

jpetitcolas:
Not sure how, but can you try adding some tests on it?

fzaninotto:
Thanks!

PR #713, opened 2015-09-30
[RFR] All fields have template

fzaninotto:
Now all field types have a `.template()` method - so the need to create a custom type becomes quite rare. 

- [x] make it work
- [x] add tests
- [x] add doc

In the process of clarifying which variables are available for templates in all situations, I added `value` everywhere, and it simplifies fieldViews a lot.

### DIFF
--- a/doc/Configuration-reference.md
+++ b/doc/Configuration-reference.md
@@ -467,7 +467,7 @@ A list of CSS classes to be added to the corresponding field. If you provide a f
             });
 
 * `template(String|Function)`
-All field types support the `template()` method, which makes it easy to customize the look and feel of a particular field, withut sacrificing the native features.
+All field types support the `template()` method, which makes it easy to customize the look and feel of a particular field, without sacrificing the native features.
 
     For instance, if you want to customize the appearance of a `NumberField` according to its value:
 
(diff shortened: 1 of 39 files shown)


fzaninotto:
Switching to RFR

fzaninotto:
Back to WIP, there are bugs.

fzaninotto:
And we're back to Ready For Review!

doc/Configuration-reference.md
```diff
@@ -466,6 +466,25 @@ A list of CSS classes to be added to the corresponding field. If you provide a f
                 return 'my-custom-css-class-for-list-header';
             });
 
+* `template(String|Function)`
+All field types support the `template()` method, which makes it easy to customize the look and feel of a particular field, withut sacrificing the native features.
```
jpetitcolas: without

doc/Theming.md
```diff
@@ -32,9 +32,31 @@ myEntity.listView().fields([
 ]);
 ```
 
+## Customizing The Template For A Given Field
+
+All field types support the `template()` method, which makes it easy to customize the look and feel of a particular field, withut sacrificing the native features.
```
jpetitcolas: without

PR #743, opened 2015-10-19
[WIP][BC BREAKS] Exclude angular from ng-admin-only package + introduce ng-admin-vendors-js and ng-admin-vendors-css packages

jeromemacias:
* [x] BC Break: `ng-admin-only.js` does not include any vendors anymore (`ng-admin.js` always includes vendors, no change here)
* [x] Move `webdriver-manager update` command to npm postinstall
* [x] Builded files for tests are now identically to committed build files.
* [x] Diable ng-animate in e2e tests to speed up them
* [x] Use `--hot` and `--inline` flags for `webpack-dev-server`
* [ ] Update README, UPGRADE and documentation

Here is the result:
image

Sinon server (used by fakerest) will now make a better job in demo as it is initialized before including vendors, ng-admin and restangular initialisation (for example, we will be able to fake an upload).

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "node": ">=4.2.0"
   },
   "scripts": {
+    "postinstall": "webdriver-manager update",
     "test": "make test"
   }
 }
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "angular-ui-bootstrap": "~1.2.1",
     "angular-ui-codemirror": "^0.3.0",
     "angular-ui-router": "^0.2.14",
-    "angularjs": "0.0.1",
     "babel": "^4.6.0",
     "babel-core": "^5.2.17",
     "babel-loader": "^5.0.0",
@@ -76,7 +75,7 @@
     "node": ">=4.2.0"
   },
   "scripts": {
-    "postinstall": "webdriver-manager update",
+    "postinstall": "if [ -z \"$npm_config_production\" ]; then webdriver-manager update; fi",
     "test": "make test"
   }
 }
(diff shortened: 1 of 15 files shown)


fzaninotto:
tests fail, needs rebase

jeromemacias:
Rebased and tests pass.

fzaninotto:
Any progress?

datnordstrom:
Is this still being worked on?

fzaninotto:
ping @jeromemacias

jaime-franco:
Hey Guys are you working on this ?

jeromemacias:
@fzaninotto Tests are green and result is what I expected (`ng-admin-only.min.js` + `ng-admin-vendors-js.min.css` size is equal to `ng-admin.min.js` size).
Is it ok for you ? If yes, I will add changes to `UPGRADE` file and update document accordingly.

fzaninotto:
This blog post is in favor of your approach instead of mine...

https://medium.com/@Rich_Harris/how-to-not-break-the-internet-with-this-one-weird-trick-e3e2d57fee28

Opinions?

package.json
```diff
@@ -76,6 +75,7 @@
     "node": ">=4.2.0"
   },
   "scripts": {
+    "postinstall": "if [ -z \"$npm_config_production\" ]; then webdriver-manager update; fi",
```
fzaninotto: Would be better suited in a `make install-test`

fzaninotto:
So after discussing it with @jeromemacias, we'll abandon the `-only` libs and focus on a Webpack experience.

mappedinn:
Which version should we install now?
Thanks,

PR #751, opened 2015-10-22
fix - JS error "values of undefined" on a missing reference in a column

### LINKED ISSUE #750: reference field column - js error on missing entry
When a reference field rendered as column, and the remote call does not have an entry for the reference, we get a JS error complaining that there's no such field 'values' of 'undefined'

Example scenario:
in a list of users, and given a secondary resource that list sign-in attempts, I want to show a column of "last sign-in".
            nga.field('id', 'reference')
                .label('Last Sign In')
                .targetEntity(userLastSignInAttempts)
                .targetField(nga.field('lastSignInDisplay'))
                .singleApiCall(function (ids) {
                    return { 'ids': ids, recentOnly: true };
                })

Some users might never have tried signing in yet, in that case the api call will not include them in the response, and the JS error will be written to console. 
the fix is to set the scope.value to '' if scope.entry is undefined (in maColumn)

I am also considering a setting on reference field to allow the user decide what value to render when there is no reference (overriding '')

I have a fix for this, pull request will follow soon.

kenegozi:
fixes #750

### DIFF
--- a/src/javascripts/ng-admin/Crud/column/maColumn.js
+++ b/src/javascripts/ng-admin/Crud/column/maColumn.js
@@ -36,7 +36,9 @@ function maColumn($state, $anchorScroll, $compile, Configuration, FieldViewConfi
             scope.datastore = scope.datastore();
             scope.field = scope.field();
             scope.entry = scope.entry();
-            scope.value = scope.entry.values[scope.field.name()];
+            scope.value = typeof scope.entry === 'undefined'
+                ? ''
+                : scope.entry.values[scope.field.name()];
             scope.entity = scope.entity();
             let customTemplate = scope.field.getTemplateValue(scope.entry);
             if (customTemplate) {


src/javascripts/ng-admin/Crud/column/maColumn.js
```diff
@@ -36,7 +36,9 @@ function maColumn($state, $anchorScroll, $compile, Configuration, FieldViewConfi
             scope.datastore = scope.datastore();
             scope.field = scope.field();
             scope.entry = scope.entry();
-            scope.value = scope.entry.values[scope.field.name()];
+            scope.value = typeof scope.entry === 'undefined'
+                ? ''
+                : scope.entry.values[scope.field.name()];
```
jeromemacias: `scope.value = scope.entry.values ? scope.entry.values[scope.field.name()] : '';` is better IMO.

kenegozi: it's not that .values is undefined, it's that .entry is undefined, so the one-liner will still end up with `Uncaught TypeError: Cannot read property 'values' of undefined`

kenegozi: as for relying on truethiness vs. checking typeof, this could turn into a holy war which is far from my intent, so whichever is consistent with the project's guidelines is fine by me.

fzaninotto:
I really don't understand why the build fails on this PR. I've restarted it three times.

Could you try rebasing on master?

kenegozi:
rebased, still same failure. I have no idea what's causing it.

fzaninotto:
OK, merging with hope that the faulty CI worker was only on your branch...

kenegozi:
Thanks.

This is so weird. I can't explain it.

lucval:
I think it might be related to this one: when using function(entry) in field's cssClasses exception "entry is undefined" is raised. I did a bit of debugging and it looks as this exception is raised before the actual data is received (in fact the cssClasses function completes without issues).

PR #756, opened 2015-10-30
[RFR] Add support for the `View.prepare()` logic

fzaninotto:
`view.prepare(Function)` adds a function to be executed before the view renders.

This is the ideal place to prefetch related entities and manipulate the DataStore. The function can be asynchronous, in which case it should return a Promise.

The `prepare` function is invoked using Angular's dependency injection system, with a context offering the following services:

- `query`: the query object (an object representation of the main request query string)
- `datastore`: where the Entries are stored. The dataStore is accessible during rendering
- `view`: the current View object
- `entry`: the current Entry instance (except in listView)
- `entries`: the current list of Entry instances (only in listView)
- `Entry`: the Entry constructor (required to transform an object from the REST response to an Entry)
- `window`: the window object. If you need to fetch anything other than an entry and pass it to the view layer, it's the only way.

Of course, regular Angular services (like Restangular) are also available.

```js
post.listView().prepare(['datastore', 'entries', 'Entry', function(datastore, entries, Entry ) {
  const authorIds = entries.map(post => post.values.authorId).join(',');
  return fetch('http://myapi.com/authors?id[]=' + authorIds)
     .then(response => response.json())
     .then(authors => Entry.createArrayFromRest(
         authors,
         [new Field('first_name'), new Field('last_name')],
         'author'
     ))
     .then(authorEntries => datastore.setEntries('authors', authorEntries));
}])
```

- [x] Add call to `prepare()` in routing
- [x] Document usage
- [x] Add working example
- [x] Add tests

### DIFF
--- a/doc/Custom-types.md
+++ b/doc/Custom-types.md
@@ -240,19 +240,19 @@ Fetching data from the `link()` function of a directive has drawbacks:
 
 The alternative is to prepare the data before the page starts rendering, to store it in the `datastore`, and to pass the `datastore` to the custom directive.
 
-Fortunately, every view has a `prepare()` method, which expects a function as parameter. This function is executed before the view is rendered. It receives the current DataStore instance (already filled with all the entries required by the view). It's the ideal place to group fetches to a remote API, and store the results in the `datastore`:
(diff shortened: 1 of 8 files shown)


fzaninotto:
switching to RFR

tal-ni:
from which version is the prepare function?

PR #758, opened 2015-11-03
[RFR] Cleanup

fzaninotto:
It's time to start preparing the 1.0 release. Since we're in 2015, let's use ES2015 when it makes sense.

Warning: extremely long diff, zero code change.


src/javascripts/ng-admin/Crud/misc/view-actions.html
```diff
@@ -1,12 +0,0 @@
-<span ng-repeat="button in buttons" ng-switch="button" class="view_actions">
```
fzaninotto: moved to the ViewActions directive

src/javascripts/test/unit/Main/component/service/config/ApplicationSpec.js
```diff
@@ -1,236 +0,0 @@
-/*global jasmine,angular,describe,it,expect*/
```
fzaninotto: Moved to admin-config (see https://github.com/marmelab/admin-config/pull/49)

Issue #730, opened 2015-10-12
Override default path


nicovak:
Hello,

I'm trying to embed ng-admin in an existing angular app.
	
	.when("/admin", {
		 templateUrl: "./partials/admin/index.html",
	});

In the view I simply added:

	<div ui-view></div>

It works but I have a not found page, I only have the left menu displayed.

How can i handle the whole ng-admin in a subpath, each link seems to be broken.

![Image of the sandbox](http://i.imgur.com/qGMHGFf.png)

You can see my work [HERE](https://github.com/nicovak/sfsandbox)

Thank you :)

fzaninotto:
Duplicate of #335 - you can't for now.

Issue #446, opened 2015-05-18
Consider using a named ui-view


sgimeno:
First of all props to all maintainers for the great job on this angular module. I'm already using it in a couple of projects.

When writing a simple Angular web app to display ng-admin views as a "backoffice" for some data source I can't build other angular-ui-router `ui-view`'s because ng-admin needs an unamed `ui-view`. Wouldn't be nice to be able to embed ng-admin views in a separated and named `ui-view`? Is there any other approaches I'm missing or is this out of the scope of the project?

So far I also found this article which encourages to use named `ui-view`'s:
http://www.toptal.com/angular-js/top-18-most-common-angularjs-developer-mistakes#common-mistake-5-not-using-named-views-with-ui-router-for-power

fzaninotto:
Hi Sergio, and thanks for your feedback.

Yes, we are aware of this issue, and using a not-main view is indeed the solution. The only problem is that it would break existing applications.

I guess the best way (and the only BC way) would be to add a second parameter to the `configure()` call, specifying the named view to use (and it would be `main` by default). 

@jpetitcolas any feedback on this?

fzaninotto:
Related to #335

jpetitcolas:
Passed parameter is already called `config`. I think we can simply add a new attribute `uiViewName` to this object, like we added the [debug](https://github.com/marmelab/ng-admin/blob/master/examples/blog/config.js#L44-L46) one recently.

Moreover, isn't the name `main` too generic? I would name it `nga-main` to prevent from collisions.

fzaninotto:
The `main` state replaces the default (nameless) ui-view. It's required for BC. 

I'm +1 with your suggestion to add a parameter to the `config` object.

sgimeno:
+1

amejias101:
+1

Siafu:
So what's the work around ? My app is currently configured as follows. 
app.js
![image](https://cloud.githubusercontent.com/assets/632024/9538673/f4de84b4-4d14-11e5-978c-310f36d0d15f.png)

admin.module
![image](https://cloud.githubusercontent.com/assets/632024/9538687/0f127cc8-4d15-11e5-9072-7845ef388ee7.png)

admin.route
![image](https://cloud.githubusercontent.com/assets/632024/9538691/1be1690a-4d15-11e5-8346-4bedd9abb8dc.png)

admin html
[...]

PR #759, opened 2015-11-03
[BC BREAK] [RFR] Rename main view to ng-admin

### LINKED ISSUE #335: Using ng-admin app inside another app?
I have built an angular app using ui-router, with 3 different abstract states for three different contexts:

* A front-office, on the `/`route
* A back-office for users, on the `/account` route
* An administration panel, on the `/admin` route

I'm not too sure about running another app just for the admin that I would have to assign to the `/admin` route in my express server, so I was hoping I could just use ng-admin as a module and assign it to my `/admin` route within the app using ui-router or something else, but I see no documentation to support that it is feasible.

So... is it? Or am I a hopeless dreamer :(

fzaninotto:
I don't think we can configure the view name, since the routing executes before we can access the configuration...

Closes #335, #446, #730

### DIFF
--- a/src/javascripts/ng-admin/Main/config/routing.js
+++ b/src/javascripts/ng-admin/Main/config/routing.js
@@ -16,17 +16,21 @@ function entryConstructorProvider() {
 
 function routing($stateProvider, $urlRouterProvider) {
 
-    $stateProvider.state('main', {
+    $stateProvider.state('ng-admin', {
         abstract: true,
-        controller: 'AppController',
-        controllerAs: 'appController',
-        templateProvider: ['NgAdminConfiguration', function(Configuration) {
-            return Configuration().layout() || layoutTemplate;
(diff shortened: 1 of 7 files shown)


src/javascripts/ng-admin/Main/config/routing.js
```diff
@@ -16,17 +16,21 @@ function entryConstructorProvider() {
...
+                controller: 'AppController',
+                controllerAs: 'appController',
+                templateProvider: ['NgAdminConfiguration', function(Configuration) {
+                    return Configuration().layout() || layoutTemplate;
+                }]
+            }
+        }
     });
 
     $stateProvider.state('dashboard', {
```
jpetitcolas: Note that we can also use dotted notation for parent views: `ng-admin.dashboard` for instance.

Issue #474, opened 2015-05-28
Template field in CSV ExportFields list


lucval:
It looks like that currently it is not supported.

I've made a patch that you can find here: https://gist.github.com/lucval/e1db83a73c90c56104bb#file-gistfile1-txt

fzaninotto:
Hi, would you like to submit a PR so that we can merge your patch, but keeping you as the author?

jpetitcolas:
Any update on it @lucval? :)

We can also make the PR ourselves, but your name won't appear in contributors list then. :(

lucval:
...finally #761

fzaninotto:
Fixed by #761

PR #761, opened 2015-11-04
Support template field in CSV ExportFields list

lucval:
This was an old fix that for some reason I forgot to pull... I saw that in the new version, this still is not fixed, therefore here it is.

Ref #474


fzaninotto:
Thanks!

PR #771, opened 2015-11-10
Possibility to add Create action Button for any entity  and fixes Bug

bananenmannfrau:
This adds the option, to add a button which allows to specify the entity for which the create view should be done.

If this functionality isnt useful for a broader audience this still fixes a bug where attrs.entityName is called when it should be attrs.entity.

### DIFF
--- a/src/javascripts/ng-admin/Crud/button/maCreateButton.js
+++ b/src/javascripts/ng-admin/Crud/button/maCreateButton.js
@@ -18,10 +18,10 @@ export default function maCreateButtonDirective($state) {
             scope.gotoCreate = () => {
                 var entityName;
                 if (attrs.force === "true"){
-                  entityName = attrs.entityName;
+                  entityName = attrs.entity;
                 }
                 else {
-                  entityName = scope.entity() ? scope.entity().name() : attrs.entityName;
+                  entityName = scope.entity() ? scope.entity().name() : attrs.entity;
                 }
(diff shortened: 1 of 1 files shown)


fzaninotto:
I don't understand. This directive already allows you to set the target entity either using an `Entity` object (`entity="entity"`), or using an entity name (`entityName="myEntity"`). So you can already do what you ask.

Also, about your bug, when does it occur?

bananenmannfrau:
I added this button to the listView listactions

'<ma-create-button "entity="price" default-values="{}" size="xs"></ma-create-button>'

same with 

'<ma-create-button entityName="price" default-values="{}" size="xs"></ma-create-button>'

When clicking on it I am just getting a 404 not found page.

I guess this is because the entity is already in the scope and therefore the custom one is not used.


Regarding the bug, that was just a misunderstanding on my side.

fzaninotto:
No, `<ma-create-button>` has isolated scope. What if you try:

```
<ma-create-button entity="entity" default-values="{}" size="xs"></ma-create-button>
or
<ma-create-button entity-name="price" default-values="{}" size="xs"></ma-create-button>
```

bananenmannfrau:
The latter one works. 

On a side note:
Is there a possibility to access the entry in the defaultvalues to pass its identifierValue?
like
    default-values="{'item' : entry().identifierValue }"

Issue #764, opened 2015-11-06
Edit UI: apply css class to whole row


dadmin-admin:
I tried to configure field following way:
````
nga.field('url').cssClasses('hidden')
````
Since css class is applied to input div, input is hidden but label is still displayed.

How to hide/apply class to whole row?

fzaninotto:
Duplicate of #755

PR #772, opened 2015-11-10
[RFR] Add the ability to set field template with label

### LINKED ISSUE #755: Hide a row from editionView depending on some conditions
Hi 

I've used ng-admin for only 2 weeks, so, please, forgive me if I've misunderstood something.

I would like to be able to hide a whole row from editionView - label+field - depending on current value of some other field.

If I set cssClasses to 'ng-hide' it hides only fields but not the label and the row still exists.

Is there any solution of my issue?

Thanks a lot!

fzaninotto:
In `showView`, `editionView`, and `creationView`, the template zone covers only the field itself - not the label. To force the template to replace the entire line (including the label), pass `true` as second argument to the `template()` call. This can be very useful to conditionally hide a field according to a property of the entry:

```js
post.editionView()
    .fields([
        nga.field('category', 'choice')
            .choices([
                { label: 'Tech', value: 'tech' },
                { label: 'Lifestyle', value: 'lifestyle' }
            ]),
        nga.field('subcategory', 'choice')
            .choices(function(entry) {
                return subCategories.filter(function (c) {
                    return c.category === entry.values.category;
                });
            })
            // display subcategory only if there is a category
            .template('<ma-field ng-if="entry.values.category" field="::field" value="entry.values[field.name()]" entry="entry" entity="::entity" form="formController.form" datastore="::formController.dataStore"></ma-field>', true),
    ]);
```

Closes #755, #764


jamespsterling:
When I try the example above, I get an infinite loop. Any idea what's causing this?

```
    submissions.editionView().fields([
        nga.field('submittedById', 'reference')
            .label('Submitted By')
            .targetEntity(user)
            .targetField(nga.field('name'))
            .editable(false)
            .template('<ma-field ng-if="entry.values.submittedById" field="::field" value="entry.values[field.name()]" entry="entry" entity="::entity" form="formController.form" datastore="::formController.dataStore">', true),
```

Error:
```
TypeError: Cannot read property 'submittedById' of undefined
```

psorowka:
@jamespsterling did you resolve this infinite loop? I am observing the same behavior.

ajarbol:
Yes the digest loop is triggered when this is used with `embedded_list` `targetFields`...

psorowka:
@ajarbol thanks for your comment! However, I just realized that my problems seem to result in the fact, that this feature hasn't actually been released to npm yet :) In the current release (0.9.1), the second parameter on `template()` is not handled and when applying an `<ma-field>` tag then, the loop is caused.

@fzaninotto this feature is already described in the docs but not yet available in the release. Do you plan to include it in an upcoming release?

fzaninotto:
Yes it will be in the next release, with no release date yet.

psorowka:
thanks. so at least this is documented allowing others to avoid this struggle :)

jamespsterling:
Is this still unreleased? Adding true as the 2nd argument to template method does nothing.

fzaninotto:
You need to use version 1.0.0-alpha4 (which is not labeled stable) to test it.

jamespsterling:
@fzaninotto great, I'll give it a shot! Any idea when 1.0.0 will be released?

PR #776, opened 2015-11-12
Fix typo in arguments to prepare in router

yarbelk:
- router's prepare injected 'entry', passed in `entries`,
 and used `entry`, changed so that they are all consistently 'entry'


fzaninotto:
Thanks!

PR #779, opened 2015-11-12
Fixed backwards scope.choices test.  Fixes #778.

### LINKED ISSUE #778: Updated to 0.9, now getting `t.choices` is not a function
Every time I switch to an edit page, I'm now getting this error:

```
Error: t.choices is not a function
r/<.compile/<.pre@http://localhost:3001/adm/js/vendor/ng-admin/build/ng-admin.min.js:15:2543

«snipped Angular call stack »

 <ma-choice-field class="ng-isolate-scope" field="::field" entry="entry" value="value">
```

I've narrowed it down to [line 36 of maChoiceField.js](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/field/maChoiceField.js#L36).  I don't know exactly what's wrong.  I think it's a mistake, because it calls the function _then_ uses the result.

Looks like the change was made in commit 0f0693647f252783d7cb4b0ecd6f1ef00c333167.

If it matters, these fields are using a custom function for the choices.  It's a bit of a hack to allow me to asynchronously load some meta data from the server, but it shouldn't affect anything.  The way it works is like this:

```js
let someChoices = [];
app.config(function(...) {
    field.choices(() => someChoices);
});

...

app.run(function ($http) {
    // load remote values
    someChoices = remoteValues;
});
```

I'd make a minimal test case, but I'm honestly trying to figure out how to do that, since you need a RESTful backend.

OverZealous:
Note: I couldn't test the code directly, because `make install` was erroring on me, and I don't have the time this second to dig into it.

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maChoiceField.js
+++ b/src/javascripts/ng-admin/Crud/field/maChoiceField.js
@@ -33,7 +33,7 @@ export default function maChoiceField($compile) {
                         refreshAttributes = 'refresh-delay="refreshDelay" refresh="refresh({ $search: $select.search })"';
                     }
 
-                    var choices = scope.choices() ? scope.choices : (field.choices ? field.choices() : []);
+                    var choices = scope.choices ? scope.choices() : (field.choices ? field.choices() : []);
                     var attributes = field.attributes();
                     scope.placeholder = (attributes && attributes.placeholder) || 'Filter values';
 


src/javascripts/ng-admin/Crud/field/maChoiceField.js
```diff
@@ -33,7 +33,7 @@ export default function maChoiceField($compile) {
                         refreshAttributes = 'refresh-delay="refreshDelay" refresh="refresh({ $search: $select.search })"';
                     }
 
-                    var choices = scope.choices() ? scope.choices : (field.choices ? field.choices() : []);
+                    var choices = scope.choices ? scope.choices() : (field.choices ? field.choices() : []);
```
fzaninotto: nah, `choices` being passed to the directive with an `&`, it's always a function. So I guess the right fix should be 

```js
var choices = scope.choices() || (field.choices ? field.choices() : []);
```

OverZealous:
Closing this, though the issue remains, as this PR is the wrong solution.

I think the right solution might be
```js
(angular.isFunction(scope.choices) && scope.choices()) ? scope.choices : (field.choices ? field.choices() : []);
```

I'm guessing that Angular is seeing the `&?` and not assigning the value if it's not set, but I honestly have no idea.  I have not seen `&?` used much.

PR #785, opened 2015-11-13
[RFR] Use ui-sref for links instead of ng-click

### LINKED ISSUE #775: Links should use `ui-sref` or `href` whenever possible to allow for open-in-new-tab
Right now, all in-app links are using `ng-click` and calling a function, which means it's not possible to open a page in a new tab or new window.

Since everything has a unique URL, it would be nice if ng-admin used `ui-sref` to navigate whenever possible, which would automatically set up things like proper `href` attributes.

(I might be able to scrape some time together to look at adding this in a PR if this is something others desire, too.)

fzaninotto:
Allows to visualize links before they are clicked, middle-click to open in a new tab, copy the link, etc.

Closes #775

### DIFF
--- a/UPGRADE-1.0.md
+++ b/UPGRADE-1.0.md
@@ -1,0 +1,27 @@
+# Upgrade to 1.0
+
+## No more `gotoDetails() and `gotoReference()` in column scope
+
+If you wrote a custom FieldView, you probably based the Link widget on the ones bundled by ng-admin:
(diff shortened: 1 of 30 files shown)


fzaninotto:
Back to WIP: referenced_list links reference the parent entity

fzaninotto:
And back to RFR

PR #788, opened 2015-11-14
remote-complete filtering

### LINKED ISSUE #787: remote-complete issue - results are not set if term does not perfectly match part of the choice
When using remote complete on choice or choices, with either reference or reference_many, the behavior is such that the returned values has to match a Contains predicate in order to work, otherwise the results do not show up, this is due to the | filter in the directive's template.

the | filter is needed for local filtering of options, but when using a remote call, the remote server is in charge of filtering, so local filtering gets in the way 

A sample scenario: when adding a reference to something with a title, the backend might treat incoming request (like `ang adm`) as prefix terms, returning `{1, "Angular Admin" }` as the result. Since `"Angular Admin").indexOf('ang adm') < 0`, the value that was returned would not show up on the dropdown, and would not be selectable.

A proposed solution (in an upcoming PR) removes the filter in case of a remote-complete instance of the field.

kenegozi:
remote-complete issue - results are not set if term does not perfectly match part of the choice

- fixes #787

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maChoiceField.js
+++ b/src/javascripts/ng-admin/Crud/field/maChoiceField.js
@@ -28,9 +28,11 @@ export default function maChoiceField($compile) {
                     });
 
                     var refreshAttributes = '';
+                    var itemsFilter = '| filter: {label: $select.search}';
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

Issue #780, opened 2015-11-12
Auto-complete dropdown menu occasionally appears at the top of the page


cfatt10:
Sometimes when searching an auto-complete dropdown menu the menu will be placed at the top of the page via CSS (position: absolute, top:-200px)

fzaninotto:
True, I've noticed it, too, that's a bug.

fzaninotto:
Unfortunately, the `position` ui-select attribute that should solve this issue isn't available in the version of ui-select we use. And upgrading ui-select is a no-no (refs https://github.com/angular-ui/ui-select/issues/1398 and https://github.com/marmelab/ng-admin/pull/869).

Kmaschta:
The PR  #1189 upgraded the version of `ui-select`.
Then, we can address the issue if someone can reproduce it.

PR #790, opened 2015-11-14
controlling choices dropdown position - workaround #780

kenegozi:
using newer ui-select, and allowing the user to pass position values to it's ui-select-choices to control the position of the dropdown. There are a few bugs in ui-select for when it opens up (e.g. when there's not enough room for it to open up), and allowing the user to set position:"down" would work around this (you can almost always scroll down, or add padding to allow for it, but not up).


mRoca:
Caution with the 0.13 ui-select branch : https://github.com/marmelab/ng-admin/pull/869#issuecomment-167577269

fzaninotto:
We've updated ui-select, it should be fixed by now on master.

PR #806, opened 2015-11-19
workaround ng 1.3.20 bug with filterless track by on choice field

kenegozi:
choice field, with remoteComplete, causes ng-admin to freak out with 'unexpected token "track"', see see https://github.com/angular-ui/ui-select/issues/1233

the workaround is to put in a noop filter, at least while 1.3 is supported.


fzaninotto:
ng-admin 1.0 will require angular 1.4. Should it solve the issue that way.

kenegozi:
I tries building with angular 1.4 (changed package.json to point to ~1.4.7) and it looks like this specific problem went away. 

We could put the workaround in place, and remove it when officially dropping 1.3 support.

As of the state of current master branch, are there any known problems with 1.4?

fzaninotto:
Fixed by #945

Issue #738, opened 2015-10-15
Export works incorrect after deleting filters and search query


AlexH7:
__Steps for reproduce:__
    1. Use some filter or search for your entities list.
    2. Delete your filter or search query after looking of new filtering results.
    3. Do the Export operation for this list and look on the results.

__Result:__
    The export results contains data from previous filtering operation.

__Expected:__
    The export results should contains data from the current list view.

lucval:
Just accidentally bumped into this issue, quite annoying indeed.
+1

lucval:
Fixed by #835

re1naldo:
I still got incorrect result when filters were deleted. This also happened when I sorted records. Data in list view was sorted properly, but the exported result (CSV) was sorted by ID (descending). Am I missing anything here?

lucval:
@re1naldo: if I understand your scenario correctly, when removing a filter your exported data is still containing (or missing) the previously filtered data?
I just tried all scenarios (i.e. data is either added, removed or both after removing a filter) and the exported CSV was always containing the correct records.

You do have a case though about the sorting order which is not fully respected by the export feature. My suggestion to you is to open a new issue and it would be nice if you could also work on a solution!

re1naldo:
After using ng-admin from master branch, the filtering feature works fine now. I will open different issue regarding data sorting.

PR #835, opened 2015-12-07
Fix ng-admin issue #738

lucval:
exportToCsv function was failing to correctly (re)load data in a new filtered page.
This was due to an incorrect search parameters given to the ReadQueries getAll function. Using $stateParams' search attribute (instead of scope.search()) solves the issue.

As a bonus a minor fix:
Use separate block to specify CSS classes in show item view.
This is required to cope with a bug found in webkit-based browsers (e.g. Chrome, Safari) which produced a corrupt highlighted text upon both double-click and triple-click selection.


lucval:
Sorry, I'll made two separate pull requests.

PR #842, opened 2015-12-10
Reload export data when list is (re)filtered

lucval:
exportToCsv function was failing to correctly reload data in a new filtered page. This was due to an incorrect search parameters given to the ReadQueries getAll function. Using $stateParams' search
attribute (instead of scope.search()) solves the issue.

Fix ng-admin issue #738.

### DIFF
--- a/src/javascripts/ng-admin/Crud/button/maExportToCsvButton.js
+++ b/src/javascripts/ng-admin/Crud/button/maExportToCsvButton.js
@@ -4,8 +4,7 @@ export default function maExportToCsvButton ($stateParams, Papa, notification, A
         scope: {
             entity: '&',
             label: '@',
-            datastore: '&',
-            search: '&'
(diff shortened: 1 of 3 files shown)


src/javascripts/ng-admin/Crud/button/maExportToCsvButton.js
```diff
@@ -4,8 +4,7 @@ export default function maExportToCsvButton ($stateParams, Papa, notification, A
         scope: {
             entity: '&',
             label: '@',
-            datastore: '&',
-            search: '&'
```
fzaninotto: You'll have to remove the param where the directive is called, in `listLayout.thml`

lucval:
You are right, forgot to push it. It is fixed now.

fzaninotto:
Thanks!

PR #843, opened 2015-12-10
Fix double-click selection bug in webkit browsers

lucval:
Use separate block to specify CSS classes in show item view.
This is required to cope with a bug found in webkit-based browsers (e.g. Chrome, Safari) which produced a corrupt highlighted text upon both double-click and triple-click selection.


fzaninotto:
Thanks!

PR #862, opened 2015-12-16
[POC] Add creation & edition events

mRoca:
Hi,

While using ng-admin, I've found two limits for my needs :

- I've not found an easy way to mark a field as invalid after API POST or PUT call (the API 400 response contains violations)
- After an entity update, if the API has updated a field (slug generation for example), the new values returned by the call are not set in the entity (there is a `transformToRest`, but no `updateToRest` method call in the [submitEdition](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/form/FormController.js#L62) method.

In order to fix this problems, we can use a `updateToRest` method [here](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/form/FormController.js#L72) and add a `handleCustomError` [here](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/form/FormController.js#L83), or we can add some events in the `FormController` and let the developer do what it wants. 

This PR is a POC, what do you think about the event usage ?

### DIFF
--- a/doc/Theming.md
+++ b/doc/Theming.md
@@ -181,3 +181,32 @@ entity.errorMessage(function (response) {
     return 'Global error: ' + response.status + '(' + response.data + ')';
 });
 ```
+
+## Events
+
+Some events are available when creating or updating an entity :
+
+- `ng-admin.creation.success`
+- `ng-admin.creation.error`
+- `ng-admin.edition.success`
+- `ng-admin.edition.error`
+
+Example :
+
+```js
+angular.module('my_module', [])
+    .run(function($rootScope) {
+        $rootScope.$on('ng-admin.edition.error', function(event, args) {
+            if (angular.isUndefined(args.response.data.violations)) {
+                return;
+            }
+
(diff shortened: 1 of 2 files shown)


fzaninotto:
Interesting approach. However, it needs more work because:

- it's hard to target a specific entity with event interceptors, and the code may live far from the entity configuration. I'd prefer something living in `admin-config`
- an event can't change the workflow of the form submission (e.g. after creation, redirect to another page than the edition page of the new entry). We should use a the return value of a function instead.
- if the event handler is asynchronous, the screen may flicker once it updates the scope. 
- I'm not very fond of event handlers updating things outside of their scope (which is the only way for a `ng-admin.edition.success` handler to update `this.$scope.entry`);

I don't have a better solution for now, but I'll be thinking about it. Anyway, thanks for taking the time to look into this.

fzaninotto:
Superseded by #898

PR #869, opened 2015-12-28
Update ui-select version

mRoca:
This PR fixes ui-select bugs by using the last version (0.13.2).

For example : the https://github.com/angular-ui/ui-select/pull/836 PR add some display bugs like https://github.com/angular-ui/ui-select/issues/1007 . The last version fixes this.

CAUTION : see the next comment

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "style-loader": "^0.12.2",
     "superagent": "^0.18.2",
     "textangular": "1.4.3",
-    "ui-select": "angular-ui/ui-select#271bf6a03078587c5afdb05f61e826573a13d348",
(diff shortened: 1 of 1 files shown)


mRoca:
The 0.13 ui-select branch is not compatible with ng-admin because of  [track by](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/field/maChoiceField.js#L45) ng-repeat expression.

fzaninotto:
Even worse: https://github.com/angular-ui/ui-select/issues/1398

fzaninotto:
Done in #1189

PR #874, opened 2015-12-31
Adds support for UTF-8 BOM

shahar3012:
This change adds the support for languages such as Hebrew and Arabic.


fzaninotto:
Doesn't work in my version of Excel (Mac, western encoding), the BOM appears in clear text in the first line. I suggest you make it an option of the `<ma-export-to-csv-button>`, disabled by default.

Also, why did you change the mime type?

fzaninotto:
No news for a month, closing.

flyseamomo:
it work in my project

PR #878, opened 2016-01-05
Fix uglifyjs error

jontonsoup:
Fix uglifyjs error when required from npm / webpack.

Uglify seems to not want to parse already minified code and hangs.


jontonsoup:
Whoops -- this error still exists, but my fix didn't work :)

PR #889, opened 2016-01-11
skip DetailLink when entry is not defined.

aitboudad:


fzaninotto:
How can entry be undefined?

aitboudad:
see https://github.com/aitboudad/ng-admin/blob/ma-column/src/javascripts/ng-admin/Crud/column/maColumn.js#L39

fzaninotto:
I guess that's when a `reference` field points to a non-existent entry, am I right?

aitboudad:
yes

src/javascripts/ng-admin/Crud/column/maColumn.js
```diff
@@ -51,12 +51,16 @@ export default function maColumn($state, $anchorScroll, $compile, Configuration,
...
             }
-            scope.detailState = getDetailLinkRouteName(scope.field, scope.entity);
-            scope.detailStateParams = {
-                ...$state.params,
-                entity: scope.entry.entityName,
-                id: scope.entry.identifierValue,
-            };
+            scope.detailState = false;
+            scope.detailStateParams = {};
+            if (angular.isDefined(scope.entry)) {
```
fzaninotto: please use `typeof scope.entry === 'undefined'` instead to avoid the dependency on the `angular` global

aitboudad: done

fzaninotto:
Thanks!

Issue #782, opened 2015-11-13
How to mark field as invalid after server validation?


dadmin-admin:
Hello. Is it possible to add server validation to the form?

Here the description of expected workflow:
1. User submits the form
2. Server validates data
3. On incorrect data server return error messages
4. Client marks fields as invalid + display error messages per field
5. On any field value change field is marked as valid and error message is hidden

fzaninotto:
That's an enhancement. And the problem is that REST doesn't standardize validation error reponses at all. So I guess the answer for that will be #777

re1naldo:
After the addition of `onSubmitError`, is it possible to display error messages below each field? In the example, you only used `notification.log('...')`, which only set an error message for the whole form.

fzaninotto:
You'll have to override the default `maField` template to display these error messages close to the field. We use angular's built-in form validation, which doesn't support this out ouf the box.

Abolfouz:
Hello,
Please, I am trying to use the onSubmitError handler but I am getting the following error:
users.creationView(...).onSubmitError is not a function

re1naldo:
Did you use ng-admin from master branch? #833

Abolfouz:
No, I have just used it. and the following error appeared 

`TypeError: t(...) is null`

Issue #777, opened 2015-11-12
Trigger custom action after submit


huvber:
I need to trigger some custom action after submit (for example return to the previous page, instead of remain in the same page). Is it possible to add a behaviour to the submit trigger?

fzaninotto:
That's not built in, it's an enhancement request.

huvber:
Thanks!

kinnrot:
+1

jamespsterling:
+1 for sure, some sort of custom callback would be awesome.

fedotxxl:
+1. Please implement it

wadjeroudi:
+1
Is there a workaround right now if you want to do something after entity changes ?

lucval:
Yes implement your own logic in the form controller (https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Crud/form/FormController.js).

It is as simple as a $state.go function call.

devguf:
+1

tarantsoff:
+1

PR #898, opened 2016-01-19
[RFR] Add edition and creation hooks

### LINKED ISSUE #805: Edition view does not load entire entity after  save changes
Lets say I have a field x that his value is calculated on server side based on field y
if I update x value via admin ui, click save, than server side changes y returns it, but the change is not reflected on ui without refreshing the page

fzaninotto:
Closes #805, #862, #782, #777.

See updated doc for details.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -4,10 +4,11 @@
 * [My API requires Authentication. How Can I Set it up?](#my-api-requires-authentication-how-can-i-set-it-up)
 * [How Can I Display A Composite Field?](#how-can-i-display-a-composite-field)
 * [How Can I Map Twice The Same Field In A List?](#how-can-i-map-twice-the-same-field-in-a-list)
+* [How Can I Handle Server-Side Validation?](#how-can-i-handle-server-side-validation?)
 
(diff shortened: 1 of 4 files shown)


mRoca:
Great job ! Thanks a lot :)

windmaomao:
When you invoke this onSubmitSuccess, can you pass in the previousState as well ?

```

            .then(entry => view.onSubmitSuccess() && this.$injector.invoke(
                view.onSubmitSuccess(),
                view,
                { previousState, $event, entity: this.entity, entry, controller: this, form: this.form, progression, notification }
            ))

```

because if we pass that, then i can do the redirect to save all list filter parameters, as 

```

        editionView.onSubmitSuccess(['previousState', 'progression', 'notification', '$state', 'entry', 'entity', function(previousState, progression, notification, $state, entry, entity) {
            // stop the progress bar
            progression.done();
            // add a notification
            notification.log(`Element #${entry._identifierValue} successfully edited.`, { addnCls: 'humane-flatty-success' });
            // redirect to the list view
            $state.go(previousState.name, previousState.params);
            // cancel the default action (redirect to the edition view)
            return false;
        }])

```

PR #904, opened 2016-01-23
[RFR] Fix select multiple fails to add choice when removing value

### LINKED ISSUE #591: Multiple references field loses choices
This happens if you edit some entity that has references saved in a reference-many-field. If you delete some choice from reference field, you can't select it back clicking, you have to go back and reenter the editing form.

For example in the demo application:

1) Go to the posts list and edit any post with tags.
screen1

2) Delete any tag. You can't select it back:
screen2

3) Sometimes i also experience this behavior, but it's not persistent. It happens when i go back to dashboard while editing via changing the url manually and pressing enter, and going back to edit form:
screen3

Anyway, if you type something, the list refreshes and everything start working fine, so i don't know if it's a bug or that's the expected behavior.

By the way, the README for ng-admin-demo tells you to go to localhost:3000 but the server.js runs on port 3001.

fzaninotto:
Closes #591

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maChoicesField.js
+++ b/src/javascripts/ng-admin/Crud/field/maChoicesField.js
@@ -34,8 +34,12 @@ export default function maChoicesField($compile) {
                     var attributes = field.attributes();
                     scope.placeholder = (attributes && attributes.placeholder) || 'Filter values';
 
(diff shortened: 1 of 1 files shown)


djhi:
:+1:

PR #906, opened 2016-01-25
add undefined filtering to remove console errors due to null or empty…

jamespsterling:
… reference on showView

Not sure if this is the correct way to suppress undefined warnings, but it works.


fzaninotto:
Duplicate of #889, which is now merged.

PR #923, opened 2016-02-08
[RFR] Fix webpack-dev-server web-socket error

vasiakorobkin:
When I'm doing `make run` -  my Chromium devTools Console got flooded by "[WDS] Disconnected!" messages sent within interval of 2 seconds.
I've investigated the issue and found out that the problem is caused by `SinonJS` which intercepts WebSocket requests to `webpack-dev-server`. This is because `webpack-dev-server` recently [moved from Socket.io to SockJS](https://github.com/webpack/webpack-dev-server/pull/302), so now WebSocket requests are sent  not to `/socket.io/` path, but to the `/sockjs-node/` path.


fzaninotto:
Thanks!

PR #933, opened 2016-02-14
[RFR] Add ability to customize CSV export format

fzaninotto:
Customize the CSV export format (quotes, delimiter, newline). The default options object is `{ quotes: false, delimiter: ",", newline: "\r\n" }`.

        listView.exportOptions({
            quotes: true,
            delimiter: ';'
        });

### DIFF
--- a/doc/reference/View.md
+++ b/doc/reference/View.md
@@ -174,6 +174,14 @@ Set the fields for the CSV export function. By default, ng-admin uses the fields
                 .stripTags(true)
         ]);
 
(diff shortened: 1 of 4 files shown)


brikou:
Very usefull, thanks!

Phocea:
Great addon. Could there also be a way to customize the csv filename, for instance to add a timestamp and not just the scope.entity.name() ?

fzaninotto:
Good idea for an enhancement.

PR #942, opened 2016-02-25
fixing frontend pagination typo from '1-10 on 15' to '1-10 of 15'

nschappl:
I noticed what I believe to be a typo in the text on the frontend which specifies what items are being shown and the total amount of items. Currently the pagination area reads something like "1-10 on 15" when I think it should be "1-10 **of** 15." I updated the frontend code and the testing files to this effect.


fzaninotto:
Thanks!

PR #969, opened 2016-03-09
[RFR] Fix Datepicker JS error

### LINKED ISSUE #961: Problem with DateField
JS error in ng-admin-demo in creation view

fzaninotto:
Closes #961


fzaninotto:
Switching to RFR

src/javascripts/ng-admin/Crud/field/maDateField.js
```diff
@@ -13,7 +13,7 @@ export default function maDateField() {
         link: function(scope, element) {
             var field = scope.field();
             scope.name = field.name();
-            scope.rawValue = scope.value;
+            scope.rawValue = scope.value == null ? new Date() : (scope.value instanceof Date ? scope.value : new Date(scope.value));
```
SebLours: A `null` value can not be displayed now with
```js
scope.value == null ? new Date()
```

fzaninotto: True, already reported in #988

Issue #851, opened 2015-12-11
[Question] Sort by reference property


romqin:
Hi,

I have a problem with the sort action (in a *listView*) with a **reference** field type. Here is my field:
```javascript
nga.field('parent', 'reference')
    .targetEntity(category)
    .targetField(nga.field('title'))
    .singleApiCall(ids => ({'id[]': ids }))
```
My API return a response like this:
```json
{
    "@id": "/categories/984c633d-c5bc-4be8-bf99-8ffd30e3a7f1",
    "title": "Child category 1",
    "parent": "/categories/ed4f8bee-30b4-4f5c-a0ec-f348d9364945",
}
```
So in the view, I have the parent.title displayed.

I would like to sort by **parent.title** but I don't succeed because the request only mention **parent**.
How can I do it ?

romqin:
On [the demo](http://marmelab.com/ng-admin-demo/#/reviews/list?sortField=reviews_ListView.product_id&sortDir=ASC), you do the same with product column.
Your request is `/reviews?range=[0,29]&sort=["product_id","ASC"]`, but in this case how do you know on which product field sort ?

fzaninotto:
This is not implemented in ng-admin. Sorting on a reference field currently sorts by the reference. 

So let's consider this an enhancement request. I'll update the title accordingly.

hauptbenutzer:
Maybe the sortField could be reused for this purpose? I've made some adjustments accordingly in https://github.com/hauptbenutzer/ng-admin/commit/044fea4846dafba851a6d4f63726821015cd02f3.

PR #976, opened 2016-03-11
Reuse sortName for sorting by arbitrary fields.

hauptbenutzer:
As mentioned in #851, this allows for sorting by arbitrary field names.

### DIFF
--- a/src/javascripts/ng-admin/Crud/list/maDatagridController.js
+++ b/src/javascripts/ng-admin/Crud/list/maDatagridController.js
@@ -64,7 +64,11 @@ export default class DatagridController {
      * @returns {String}
      */
     getSortName(field) {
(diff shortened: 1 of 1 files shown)


src/javascripts/ng-admin/Crud/list/maDatagridController.js
```diff
@@ -64,7 +64,11 @@ export default class DatagridController {
      * @returns {String}
      */
     getSortName(field) {
-        return this.$scope.name ? this.$scope.name + '.' + field.name() : field.name();
+        var sortFieldName = field.name();
+        if(field.sortField) {
```
fzaninotto: where does this come from? There is no such method as `sortField` in `Field`

Issue #372, opened 2015-03-24
Remove Column from sorting


sureshvarman:
Is there any options to remove a particular column from sorting

fzaninotto:
Nope, not at the moment. What is your use case?

sureshvarman:
my requirement is simple that i don't want sorting for a column which is a template column

fzaninotto:
One may want to allow sorting by template, since the sort param can be tweaked in a request interceptor... We'll leave it like that.

0xDCA:
I believe this should be supported. In my case, the backend does not allow to sort by some fields, so trying to click on the column header just shows an error.

It would be cool if we could explicitly disable sorting for some fields

thachp:
I am open to debate on how we can best implement this enhancement.  Though, I think the following adjustments would do. 

@fzaninotto  - Can you commit these changes part of the core codes?  I am dealing with the same problem as @0xDCA.  

In admin-config, field.js

    // https://github.com/marmelab/admin-config/blob/master/lib/Field/Field.js
    add to class field attributes
    this._disableSort = false; // line 9

    // Add a set method and a get method.    

    isDisableSort() {
        return this._disableSort;
    }

    disableSort() {
        this._disableSort = true;
        return this;
    }

Update ng-repeat in grid template in src/javascripts/ng-admin/Crud/list/maDatagrid.js

    <th ng-repeat="field in fields() track by $index"
    ng-class="field.getCssClasses()" class="ng-admin-column-{{ ::field.name() }} ng-admin-type-{{ ::field.type() }}">

        // disable sort
        <a ng-if="field.isDisableSort() == false" ng-click="datagrid.sortCallback(field)">
            <span class="glyphicon {{ datagrid.sortDir === 'DESC' ? 'glyphicon-chevron-down': 'glyphicon-chevron-up' }}" ng-if="datagrid.isSorting(field)"></span>
            {{ field.label() }}
        </a>

        // remove ng-click event and replace a tag with span.    
        <span ng-if="field.isDisableSort() == true">
            {{ field.label() }}
        </span>
    </th>

seasick:
Use-case for me is, that I don't want to sort fields which are references. The displayed values will not appear sorted in most cases, because its sorted by the underlying value and not by the displayed value (e.g. sort messages by user-id, but display user-name from reference).

RichardBradley:
[...]

PR #984, opened 2016-03-17
#372 allow sortable(false) on fields in list view

RichardBradley:
Fixes https://github.com/marmelab/ng-admin/issues/372

(UPDATE: please ignore the rest of this comment, it refers to an old version of this PR. Jump straight to [the comment starting "OK, I have updated this pull request to include the necessary changes"](#issuecomment-205234237))

As discussed on that thread, this PR is much easier if I don't have to also modify `admin-config`.
However, it does look rather out of place next to all the fields with function getters & setters, so maybe a version which spans the two respositories would be better.

This version works for me though.

(Another benefit of the getter/setter approach is that it allows a "builder" style. This version is awkward to use:

```
articles.listView()
   .fields(
       nga.field('id'),
       function() {
            var headline = nga.field('headline');
            // Headline is an "analyzed" string field, so cannot be sorted in ES
            // https://www.elastic.co/guide/en/elasticsearch/guide/current/multi-fields.html
            headline.disableSort = true;
            return headline;
        }(),
        nga.field('publication_time', 'datetime'),
        nga.field('type')
    ])
```

whereas the getter/setter version would look like

```
articles.listView()
   .fields(
       nga.field('id'),
       nga.field('headline')
            // Headline is an "analyzed" string field, so cannot be sorted in ES
            // https://www.elastic.co/guide/en/elasticsearch/guide/current/multi-fields.html
            .disableSort(true),
        nga.field('publication_time', 'datetime'),
        nga.field('type')
    ])
```

)

### DIFF
--- a/doc/reference/Field.md
+++ b/doc/reference/Field.md
@@ -53,7 +53,11 @@ Create a new field of the given type. Default type is 'string', so you can omit 
 Define the label of the field. Defaults to the uppercased field name.
 
 * `editable(boolean)`
-Define if the field is editable in the edition form. Usefult to display a field without allowing edition (e.g for creation date).
+Define if the field is editable in the edition form. Useful to display a field without allowing edition (e.g for creation date).
+
+* `sortable(boolean)`
+Define if the field is sortable in the list view (default `true`).
+(See ["Sort Columns and Sort Order"](../API-mapping.md#sort-columns-and-sort-order) for a discussion of how to integrate `ng-admin` sorting with your REST backend.)
 
 * `order(number|null)`
 Define the position of the field in the view.
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "git://github.com/marmelab/ng-admin.git"
   },
   "devDependencies": {
-    "admin-config": "^0.11.0",
+    "admin-config": "^0.12.0",
     "angular": "~1.4.8",
     "angular-mocks": "~1.4.8",
     "angular-numeraljs": "^1.1.6",
--- a/doc/API-mapping.md
+++ b/doc/API-mapping.md
@@ -205,6 +205,8 @@ myApp.config(['RestangularProvider', function(RestangularProvider) {
 }]);
 ```
 
+(If a column is not sortable by your backend, e.g. a computed column, you can disable sorting per-column using [`Field.sortable(false)`](reference/Field.md#general-field-settings).)
+
 ## Filtering
 
 All filter fields are added as a serialized object passed as the value of the `_filters` query parameter. For instance, the following `filterView()` configuration:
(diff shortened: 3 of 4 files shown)


RichardBradley:
Taking a step back, why is `admin-config` in a separate git repos to `ng-admin`?
It would simplify the project a lot to merge them together; they are already very tightly coupled, so having them in separate repositories just adds administrative overhead.

fzaninotto:
`admin-config` is also used by `react-admin`, that's why it's independent. 

Please update `admin-config` with a related PR to use a more usual function syntax.

Fr4ncx:
** .disableSort(true),** Not work in my panel, why?

RichardBradley:
> .disableSort(true), Not work in my panel, why?

Because this enhancement has not yet been implemented.
The issue is still marked "open" at the top -- subscribe to #372 for updates and try again when it is marked "closed".

RichardBradley:
> Please update admin-config with a related PR to use a more usual function syntax.

Is there a guide on how best to do this?
I'll make the changes to `admin-config`; do I then need to bump its version and update the dependency reference in [`ng-admin/package.json`](https://github.com/marmelab/ng-admin/blob/master/package.json)? How will that work if the PR to `admin-config` has not been pushed yet?

RichardBradley:
OK, I have updated this pull request to include the necessary changes to `admin-config`. Please could you take another look and merge if you are happy?

This now depends on https://github.com/marmelab/admin-config/pull/65 -- please merge and release that first, or this will fail. 

I have bumped the `admin-config` version and increased the version required by `ng-admin` accordingly.

I have not included changes to `build/`, as requested in the README

RichardBradley:
(The CI build has failed because it cannot find the not-yet-published version of `admin-config` on which this change depends. This is inevitable with the current git repos structure, as briefly discussed above.)

RichardBradley:
I have changed the syntax from the proposed "disabledSort(true)" to "`.sortable(false)`", as it better matches the other Field properties like "editable"

doc/reference/Field.md
```diff
@@ -5,21 +5,21 @@ A field is the representation of a property of an entity.
 * [General Field Settings](#general-field-settings)
```
fzaninotto: please revert these changes, they are not related to this PR, and break the ng-admin book on GitBook (i.e. the official doc)

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "git://github.com/marmelab/ng-admin.git"
   },
   "devDependencies": {
-    "admin-config": "^0.11.0",
+    "admin-config": "^0.11.1",
```
fzaninotto: I have released 0.12.0

As a side note, 0.0.x is for bugfixes, 0.x is for new features. `.sortable()` is a new feature. That's called semver.

doc/reference/Field.md
RichardBradley: OK, will do.

The links are currently broken on GitHub, see e.g. https://github.com/marmelab/ng-admin/blob/master/doc/reference/Field.md#-date-field-type

They do work on the "GitBook" version, you're right: http://ng-admin-book.marmelab.com/doc/reference/Field.html#-date-field-type

It's a shame that the two differ. I hadn't spotted the "GitBook" version; I'd been referring to the documentation on GItHub directly up until now. If the GitBook version is the official one, it should win, of course.

package.json
RichardBradley: Thanks, I'll update.

RichardBradley:
I have pushed a new version with the requested changes (and rebased to account for other changes in the meantime).

fzaninotto:
This PR is OK, just needs rebasing

fzaninotto:
Superseded by #1260

PR #996, opened 2016-03-22
Update API-mapping.md

albogdano:
Fixed method names to be lowercase, as required by Restangular. Added example for date formatting.


fzaninotto:
Thanks!

PR #1000, opened 2016-03-24
WIP: Bumping angular version from 1.4 to 1.5

gorghoa:
Hi,

I’d like to use the new “component” angular method from the 1.5 version in one of my ng-admin configurations. So I wanted to test if ng-admin would be compatible with this angular 1 version.

note : I get a core dump with wepback on my local install (linux), so unable to run the tests|build|run tec. I reckon travis will run the tests for me… :-/ It seems to be the “extract-text-webpack-plugin” which causses it.

a+,


gorghoa:
damn it. Ok, i’ll dig in my segfault and have an up and running env…

gorghoa:
I’ve found that the issue is upstream, see angular-ui/ui-select#1522

Hope it won’t bubble up to angular itself…

gorghoa:
need for update dependencies  status: 

* angular-ui/ui-select#1533

Simperfit:
This one needs a rebase since there is alot of commits ;).

Phocea:
Since we are now using angular 1.6.1. I guess this can be closed?

fzaninotto:
yep, closing.

gorghoa:
congrats for admin-on-rest by the way ;)

PR #1001, opened 2016-03-24
Document setting the type for the field

geecu:
Without the type, the AmountFieldView is not used,
ng-admin uses the inherited NumberFieldView

### DIFF
--- a/doc/Custom-types.md
+++ b/doc/Custom-types.md
@@ -171,6 +171,17 @@ myApp.config(['FieldViewConfigurationProvider', function(fvp) {
 }]);
 ```
 
+To use this field view, alter the field class and set its `_type` property:
+```js
+class AmountField extends NumberField {
+    constructor(name) {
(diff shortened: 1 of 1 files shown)


doc/Custom-types.md
```diff
@@ -171,6 +171,17 @@ myApp.config(['FieldViewConfigurationProvider', function(fvp) {
 }]);
 ```
 
+To use this field view, alter the field class and set its `_type` property:
+```js
+class AmountField extends NumberField {
+    constructor(name) {
+        super(name);
+        this._type = 'amount';
```
fzaninotto: please move that to line 90, where it's more logical.

geecu: Done.

I thought the intention of the documentation was to show that the field view definition was optional, that's why I didn't add it to line 90 in the first place.

fzaninotto:
Thanks!

PR #1033, opened 2016-04-15
Updates Field.md Anchor Tags

Dan-Nolan:
Anchor tags were prefixed with a hyphen, not working properly.


Dan-Nolan:
Actually, it's broken on Github but it seems like the hyphen was intended for the live site. Closing the PR.

PR #1043, opened 2016-04-20
Error in documentation

schmittr:
The sentence `Since there a are two fields names user_id, it's a blocker.` seems to be wrong ;)


jpetitcolas:
Thanks! :)

Issue #808, opened 2015-11-21
Fields marked as not editable should not be sent to the server on save


OverZealous:
The issue is we have a bunch of embedded documents that aren't practical to edit from the Admin, so we usually perform some modification via `.map`, then render using either `'json'` or a `'template'`.

However, if the `.map` changes the value in any way, this is sent back to the server on save, even though the field is not editable, resulting in corrupted data.  This is pretty unexpected, since the field is not editable, yet the value is being edited (via map).

I'm currently using a few techniques to prevent the issue, but it's pretty tricky, especially when embedded objects are being flattened, so you can't just map them to a fake field.

Alternatively, maybe we need a way to flag fields that shouldn't be sent back.

fzaninotto:
You're right, marked as bug.

Kmaschta:
@jpetitcolas explained why we can't address this issue in the PR #1067 

>These changes may have some drawbacks. Indeed, some API may expect all the entity fields in the request payload, even the non-editable ones. Yet, as far as I understand your PR, you do not send non editable fields anymore.
>
>That's a big BC break, and I am not sure we handle it properly. A better way would be to retrieve the original entry values and to set them back to the API.
>
>Let's postpone it to another version than the 1.0.

PR #1067, opened 2016-05-16
fixing issue #808

maksimbormot:


maksimbormot:
@fzaninotto  Please check the error (https://travis-ci.org/marmelab/ng-admin/builds/130531626#L879)

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 dist
+.idea
```
fzaninotto: please include this in your *user* gitignore, this has nothing to do in a *project* gitignore

jpetitcolas:
These changes may have some drawbacks. Indeed, some API may expect all the entity fields in the request payload, even the non-editable ones. Yet, as far as I understand your PR, you do not send non editable fields anymore.

That's a big BC break, and I am not sure we handle it properly. A better way would be to retrieve the original entry values and to set them back to the API. 

Let's postpone it to another version than the 1.0.

Issue #899, opened 2016-01-19
maDateField doesn't support changing value


fedotxxl:
`value` of `maDateField` may be changed by external function (let's assume that I've got custom button wich changes entry value). 

expected: `maDateField` should display new value
result: `maDateField` displays old value

fzaninotto:
true, maDateField should watch changes on `value` to keep two way data binding. Would you like to work on a patch?

ifokeev:
need it too :(

PR #1072, opened 2016-05-18
Fixing issue 899

maksimbormot:
Fixes https://github.com/marmelab/ng-admin/issues/899 

Added watcher to `value`.
After the change should be applied via `$apply` or better via `triggerHandler('change')`.

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maDateField.js
+++ b/src/javascripts/ng-admin/Crud/field/maDateField.js
@@ -28,7 +28,6 @@ export default function maDateField() {
             scope.v = field.validation();
             scope.isOpen = false;
             var input = element.find('input').eq(0);
-
             var attributes = field.attributes();
             for (var name in attributes) {
                 input.attr(name, attributes[name]);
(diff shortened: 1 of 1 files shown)


src/javascripts/ng-admin/Crud/field/maDateField.js
```diff
@@ -17,6 +17,9 @@ export default function maDateField() {
             scope.$watch('rawValue', function(rawValue) {
                 scope.value = field.parse()(rawValue);
             });
+            scope.$watch('value', function(rawValue) {
```
fzaninotto: should be `value`

src/javascripts/ng-admin/Crud/field/maDateField.js
```diff
@@ -17,6 +17,9 @@ export default function maDateField() {
...
+                scope.rawValue = scope.value == null ? null : (scope.value instanceof Date ? scope.value : new Date(scope.value));
```
fzaninotto: you should not use `scope.value` there, but the `value` parameter

Kmaschta:
Super-seeded by #1196

Issue #1078, opened 2016-05-20
Typo in Documentation


cfatt10:
In the Field Configuration, under 'choice and choices field types', the word 'literal' is misspelled.

> When given an array, each choice must be an object **litteral** with both a value and a label.

bestori:
see [PR](https://github.com/marmelab/ng-admin/pull/1081)  :)

fzaninotto:
Fixed by #1081

PR #1081, opened 2016-05-23
[FIX] wording typo (Typo in Documentation #1078)

bestori:
Good catch, cfatt10 :)


bestori:
CI fails have nothing to do with commit..

cfatt10:
Thanks for making the pull request! It looks like you actually found another spot where 'literal' was misspelled. The original sentence was 

> When given an array, each choice must be an object litteral with both a value and a label.
And the sentence you found is

> Symmetrically, passing an object litteral as argument to the setter methor `view.fields(fields)` is deprecated ;
See issue #1078 for location of first sentence

bestori:
Hehe ok, fixing that one as well :) Any idea why checks are failing?? it appears as though it's the status of the current version...?

fzaninotto:
Thanks! I don't understand why tests fail on travis, it's unrelated to your change.

bestori:
Thanks! 👍  :)

PR #1086, opened 2016-05-23
correct anchor links

schmijos:
How could that happen in the first place?


fzaninotto:
It's on purpose, that's how the GitBook parsed creates ids, see for instance http://ng-admin-book.marmelab.com/doc/reference/Field.html#-number-field-type

PR #1092, opened 2016-05-26
[RFR] Show sort column in datagrid even with default sort

### LINKED ISSUE #1076: Default sort is not visible in column header (no chevron)
I have listView().sortField('somefield').sortDir('ASC'), and view is sorted as expected, except there is no chevron displayed next to somefield title in table header.

And, probably related, if using ASC direction, when column name first pressed, list gets sorted again in same sort (ASC) as default, so there is effectively no change except chevron gets displayed.

This issue is also visible in Posters Galore demo, click on reviews, by default they are sorted by date descending (but chevron not visible), or

fzaninotto:
Closes #1076

image


RichardBradley:
Thanks!

PR #1099, opened 2016-05-31
Update ui-select version

### LINKED ISSUE #224: json column type
Not for all users, but very handy nonetheless... Should only serialize/deserialize on top of a text field.

rigobcastro:
This update fixes "[ui.select:transcluded] Expected 1 .ui-select-match but got '0'. error


fzaninotto:
Duplicate of #869, please comment there.

Issue #1129, opened 2016-06-16
"400" error is ignored when deleting or bulk deleting items


RichardBradley:
On the current version, if the REST backend rejects a "DELETE" request from `ng-admin` no error is displayed to the user.

There is code which appears to attempt to show an error, so I think this is a regression.
I don't see any tests (unit or e2e) for this functionality.

Repro steps:

```
Given an entity where the REST backend will return 400 "cannot delete due to xyz" for a "DELETE" request
When I view the "edit" page for that entity
And I click the "delete" button
And I click "Yes" to "are you sure?"
Then I should see an error banner including "cannot delete due to xyz"
 # but I see a success banner

Given an entity where the REST backend will return 400 "cannot delete due to xyz" for a "DELETE" request
When I view the "list" page for that entity tytpe
And I select that entity
And I click the "delete" button
And I click "Yes" to "are you sure?"
Then I should see an error banner including "cannot delete due to xyz"
 # but I see a success banner
```

RichardBradley:
There are three bugs at play here, I think:
 1. The code uses "|" where it should use "||" [in a few places](https://github.com/marmelab/ng-admin/blob/c9dc3efdba38f8953535f2b6da715693f97c19c1/src/javascripts/ng-admin/Crud/delete/DeleteController.js#L46)
 2. The code uses angular-translate wrong if a custom error message is used as the translate key, due to https://github.com/angular-translate/angular-translate/issues/1516
 3. The "batch" delete does not attempt to check the individual success / failure results, but only checks whether the batch request succeeded. So even if there is only 1 item in the batch, it will look like a success to ng-admin

Phocea:
#1059 
#1036

Phocea:
#1059 #1036

RichardBradley:
Thanks, yes, looks like a duplicate of those two. I've commented there.

jpetitcolas:
Closing as duplicate.

PR #1130, opened 2016-06-16
Fix error messages on failed deletes, updates

RichardBradley:
This is a fix for items 1 & 2 described at https://github.com/marmelab/ng-admin/issues/1129#issuecomment-226500647

It does not fix item 3.

New test results:

```
Given an entity where the REST backend will return 400 "cannot delete due to xyz" for a "DELETE" request
When I view the "edit" page for that entity
And I click the "delete" button
And I click "Yes" to "are you sure?"
Then I should see an error banner including "cannot delete due to xyz"
 # this passes now (and fails before these changes)

Given an entity where the REST backend will return 400 "cannot delete due to xyz" for a "DELETE" request
When I view the "list" page for that entity tytpe
And I select that entity
And I click the "delete" button
And I click "Yes" to "are you sure?"
Then I should see an error banner including "cannot delete due to xyz"
 # this still fails due to item #3
 # I see a success banner
```

### DIFF
--- a/src/javascripts/ng-admin/Crud/delete/BatchDeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/BatchDeleteController.js
@@ -30,11 +30,13 @@ export default class BatchDeleteController {
                 $translate('BATCH_DELETE_SUCCESS').then(text => notification.log(text, { addnCls: 'humane-flatty-success' }));
             })
             .catch(error => {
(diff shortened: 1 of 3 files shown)


jpetitcolas:
Thanks for your PR. Just rebasing it and fixing tests in a new one to be able to merge it today (we plan to release a new alpha later today). :)

jpetitcolas:
It seems these changes are already on master. Closing the issue.

@Phocea: indeed, it sounds like a good idea, yet let's apply it later. Short term objective is to release 1.0 version. :)

PR #1140, opened 2016-06-28
[RFR] Display progress bar when deleting

### LINKED ISSUE #1109: deletionView doesn't have progress
Since server delete request can take some time on slow servers/links/etc... I think deletionView should also have progress indicator.

fzaninotto:
Closes #1109

### DIFF
--- a/src/javascripts/ng-admin/Crud/delete/DeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/DeleteController.js
@@ -38,9 +38,8 @@ export default class DeleteController {
                     return this.$state.go(this.$state.get('list'), angular.extend({
                         entity: entityName
                     }, this.$state.params));
-                } else {
-                    return this.back();
                 }
+                return this.back();
             })
             // no need to call progression.done() in case of success, as it's called by the view dislayed afterwards
             .then(() => $translate('DELETE_SUCCESS'))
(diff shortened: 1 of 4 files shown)


jpetitcolas:
It would be great to test that `progression.start()` and `progression.done()` are called correctly in the controller (in both case of success and case of error).

jpetitcolas:
:+1:

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -27,23 +28,26 @@ export default class DeleteController {
...
         return this.WriteQueries.deleteOne(this.view, this.entityId)
             .then(() => this.previousStateParametersDeferred.promise)
             .then(previousStateParameters => {
                 // if previous page was related to deleted entity, redirect to list
                 if (previousStateParameters.entity === entityName && previousStateParameters.id === this.entityId) {
-                    this.$state.go(this.$state.get('list'), angular.extend({
+                    return this.$state.go(this.$state.get('list'), angular.extend({
                         entity: entityName
                     }, this.$state.params));
                 } else {
```
RobinBressan: You don't need an `else` statement here

fzaninotto:
@jpetitcolas this can only be done in an e2e test (because the `done()` is called in the `resolve` phase of the next route), and is very hard to test due to timing issues. 

It is left as an exercise for the reader.

fzaninotto:
review applied

PR #1149, opened 2016-07-11
Don’t show asterisk if required field is a boolean

### LINKED ISSUE #928: Required boolean fields should not be labeled with an asterisk
An asterisk is added to the names of mandatory fields. However, required boolean fields are represented by checkboxes, and thus are already "mandatory", since there's no possible way not to fill them out.
Having them labeled with asterisks might convey a wrong idea to the user: that the fields HAVE to be ticked (as it "I agree to terms and conditions").

stevehalford:
Fixes #928


jpetitcolas:
:+1: 

Can you add a unit test? :)

Issue #791, opened 2015-11-14
Add Tooltip/Popover for field configuration


shahar3012:
Hi,

It would be great if you could add a tooltip / popover option for field configuration in the upcoming version.

Something like this:

![](http://jquer.in/wp-content/uploads/2015/01/validation-bootstrap-tooltip.jpg)
or:

![](http://i.stack.imgur.com/88hPS.png)

See: 
[Bootstrap Tooltips](http://getbootstrap.com/javascript/#tooltips)
[Bootstrap Popovers](http://getbootstrap.com/javascript/#popovers)

fzaninotto:
Neat idea. In the meantime, don't forget that you can use field templates to customize the fields at will.

shahar3012:
I think that overriding every field with a template is too much work.
Please provide an update on the status of this issue development.
Thank you.

fzaninotto:
There is no update on the status of this issue development, because this is an open-source project where contributors add features when they actually need it, and on their free time. You're welcome to give a hand, too.

RichardBradley:
I thought I wanted this UI, so I wrote a directive override to achieve it. (details below) 

On reflection, I think that this UI isn't great, even though it looks pretty on first glance.
It's fiddly to use on a desktop client; very difficult to use on mobile clients and impossible to use on a screen reader. I think that the UI proposed on #1150 is much more user-friendly, even if it perhaps isn't quite as pretty.

Here's the code I added to create these "?" icons on the current version of `ng-admin`: http://embed.plnkr.co/xQdRTUphBC3854a4VD1k/ (click "edit" on a user to see the "Father's last name only" hint)

jpetitcolas:
Perhaps should we just add a `hint` option instead, which would display a light text just below the field? Something like:

![image](https://cloud.githubusercontent.com/assets/688373/19030070/b2937820-894b-11e6-8944-e0902f69c0c8.png)

(Screenshot taken from https://github.com/callemall/material-ui/issues/2474)

RichardBradley:
@jpetitcolas -- yes, that's pretty much how the #1150 UI looks

PR #1150, opened 2016-07-12
Show help text alongside field if specified

stevehalford:
Adds help text or a description alongside fields (write only)

Related: https://github.com/marmelab/admin-config/pull/72

The directives template can be overridden to achieve this #791

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "git://github.com/marmelab/ng-admin.git"
   },
   "devDependencies": {
-    "admin-config": "^0.11.0",
+    "admin-config": "stevehalford/admin-config#field-descriptions",
     "angular": "~1.4.8",
     "angular-mocks": "~1.4.8",
     "angular-numeraljs": "^1.1.6",
--- a/src/javascripts/ng-admin/Crud/misc/maHelpText.js
+++ b/src/javascripts/ng-admin/Crud/misc/maHelpText.js
@@ -1,0 +1,20 @@
+/**
+ * Help text for a form field.
+ *
+ * @example <ma-help-text field="field"></ma-help-text>
+ */
+export default function maHelpText() {
+    return {
+        scope: {
+            'field': '&'
+        },
+        restrict: 'E',
+        link: function(scope, element) {
+            var field = scope.field();
+            scope.helpText = field.helpText();
+        },
+        template: '<small ng-if="helpText" class="help-block">{{ helpText }}</small>'
+    };
+}
+
+maHelpText.$inject = [];
--- a/src/javascripts/ng-admin/Crud/CrudModule.js
+++ b/src/javascripts/ng-admin/Crud/CrudModule.js
@@ -81,6 +81,7 @@ CrudModule.directive('maViewBatchActions', require('./button/maViewBatchActions'
 CrudModule.directive('maShowItem', require('./show/maShowItem'));
 CrudModule.directive('maViewActions', require('./misc/ViewActions'));
 CrudModule.directive('compile', require('./misc/Compile'));
+CrudModule.directive('maHelpText', require('./misc/maHelpText'));
 
 CrudModule.config(require('./routing'));
 CrudModule.config(require('./config/factories'));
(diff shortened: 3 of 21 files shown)


RichardBradley:
I think this is a great addition.

I was wondering about screen readers and whether the help text ought to be linked to the input, e.g. by a "label for". This page suggests using "`aria-labelledby`" instead: http://www.html5accessibility.com/tests/mulitple-labels.html

RichardBradley:
I'm using this on my ng-admin powered site, and it works great, thanks.

I've extended the code so that you can use HTML markup and/or Angular markup in the help text where necessary. Can I create a PR for a PR? Or should I create a separate PR that includes this code?

I have some fields which change their meaning based on the "type" of the row, and I wanted the help-text to vary. You can use it like:

```
  myEntity.editionView()
     .fields(
        nga.field('type', 'choice')
                    .choices([ .. ]),
        nga.field('searchQuery')
                    .label("Search Query")
                    .helpText(angular.element(
                        "<small ng-if=\"entry.values.type == 'live_search_feed'\" class='help-block'>" +
                        "  This is both the \"view all\" link and the " +
                        "search which will be used to fill the product list." +
                        "</small>" +
                        "<small ng-if=\"entry.values.type == 'curated_list'\" class='help-block'>" +
                        "This will be used for the \"view all\" link. (Optional)" +
                        "</small>" +
                        "<small class='help-block'>Remember to filter the search if you want only Foo" +
                        " or only Bar typed products (usually you do; it looks neater).</small>")),
```

My changes are at: https://github.com/RichardBradley/ng-admin/commit/9b3692241ae3a38eeebbb634b8fd30bdf2a94c01
(you'll need to cherry-pick that commit; it's on a branch which contains other stuff).

d1b1:
Any progress on this. Would love to have this feature.

stevehalford:
Just cherry-picked your commit @RichardBradley

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "git://github.com/marmelab/ng-admin.git"
   },
   "devDependencies": {
-    "admin-config": "^0.11.0",
+    "admin-config": "stevehalford/admin-config#field-descriptions",
```
jpetitcolas: Use new admin-config release once your https://github.com/marmelab/admin-config/pull/72 PR has been released.

src/javascripts/ng-admin/Crud/misc/maHelpText.js
```diff
@@ -0,0 +1,32 @@
...
+ * @example <ma-help-text field="field"></ma-help-text>
+ *
+ * @param field.helpText()  If a string, it will be displayed as text to the user.
+ *                          If an `angular.element`, it will be compiled and displayed.
+ *                          You can use "entry.values[x]" to check the current value
+ *                          of property "x" (e.g. if you want to give different hints)
+ */
+export default function maHelpText($compile) {
+    return {
+        scope: true,
```
jpetitcolas: Please add a comment to tell that you need parent scope in case of it is a custom directive.

src/javascripts/ng-admin/Crud/misc/maHelpText.js
```diff
@@ -0,0 +1,32 @@
...
+        restrict: 'E',
+        link: function(scope, element) {
+            var field = scope.field;
+
+            var helpText = field && field.helpText();
+            if (helpText && typeof helpText === "object" && helpText.contents) {
```
jpetitcolas: It should be the default case. First, I would test if helpText is a string, then a function, then consider it is a custom directive.

src/javascripts/ng-admin/Crud/misc/maHelpText.js
```diff
@@ -0,0 +1,32 @@
...
+                // looks like an angular.element:
+                element.append(helpText.clone());
+                $compile(element.contents())(scope);
+            } else {
+                // treat as plain text:
+                scope.helpText = helpText;
+            }
+        },
+        template: '<small ng-if="helpText" class="help-block">{{ helpText }}</small>'
```
jpetitcolas: Can you rename class to `ma-help-text`?

jpetitcolas:
Can you add some unit tests for your directive?

jpetitcolas:
:+1: 

Please rebase this PR to ensure tests are fine. :)

RichardBradley:
(If you're making changes here, it would also be nice to update the docs to explain how to use the non-string case of helpText. Maybe add the example from my comment above starting "You can use it like:" to Field.md, if it still works after jpetitcolas's changes.)

PR #1170, opened 2016-08-03
#1158 allow full-row field templates in embedded view

### LINKED ISSUE #1158: field.template(..., true) doesn't work in Embedded Entities view
When using the "[Embedded Entities](http://ng-admin-book.marmelab.com/doc/Relationships.html#embedded-entities)" edit form, the `field.template(x)` function works fine to set a custom value template, but the `field.template(x, true)` function appears to have no effect.

## To reproduce:
  *  open this Plunkr demo: http://plnkr.co/edit/YeOBmuIqqAGEcBGINtIw?p=preview
  * click on "John" on the dashboard
  * click "add new children"
  * I have added the following fields to "user" and "child" (see script.js lines 45 - 58)):

            nga.field('custom value template')
              .template("custom value text"),

            nga.field('custom row template')
              .template("custom row text", true),

## Expected behaviour:

I'd expect both the "user" form and the "child" form to have:
 * one row which has a label "custom value template" and a value "custom value text"
 * one row with just the text "custom row text" (no label)

## Observed behaviour:

The user form has the above behaviour, but the child form has:
 * one row which has a label "custom value template" and a value "custom value text"
 * one row with a label "custom row template" and a text box

RichardBradley:
This fixes #1158

### DIFF
--- a/doc/Relationships.md
+++ b/doc/Relationships.md
@@ -223,6 +223,11 @@ When displaying a post:
 
 ![embedded_list in editionView](images/embedded_list_in_editionView.png)
 
+Caution: If you want to use [`Field.template()`](./Theming.md#customizing-the-template-for-a-given-field)
+to customise the view of an embedded field, be aware that the default "<ma-field>" template string takes
+different attribute values in embedded view to the regular view.
+See the [maEmbeddedListField source](../src/javascripts/ng-admin/crud/field/maEmbeddedListField.js).
+
 ## Embedded entity
 
 APIs may render a many-to-one relationship by embedding the related entry in the main entry:
--- a/src/javascripts/ng-admin/Crud/field/maEmbeddedListField.js
+++ b/src/javascripts/ng-admin/Crud/field/maEmbeddedListField.js
@@ -59,7 +59,7 @@ export default function maEmbeddedListField() {
         <div class="remove_button_container">
             <a class="btn btn-default btn-sm" ng-click="remove(entry)"><span class="glyphicon glyphicon-minus-sign" aria-hidden="true"></span>&nbsp;<span translate="REMOVE"></span></a>
         </div>
-        <div class="form-field form-group" ng-repeat="field in ::fields track by $index">
+        <div ng-repeat="field in ::fields track by $index" compile="::field.getTemplateValueWithLabel(entry)">
             <ma-field field="::field" value="entry.values[field.name()]" entry="entry" entity="::targetEntity" form="formName" datastore="::datastore()"></ma-field>
         </div>
         <hr/>


src/javascripts/ng-admin/Crud/field/maEmbeddedListField.js
```diff
@@ -59,7 +59,7 @@ export default function maEmbeddedListField() {
         <div class="remove_button_container">
             <a class="btn btn-default btn-sm" ng-click="remove(entry)"><span class="glyphicon glyphicon-minus-sign" aria-hidden="true"></span>&nbsp;<span translate="REMOVE"></span></a>
         </div>
-        <div class="form-field form-group" ng-repeat="field in ::fields track by $index">
+        <div ng-repeat="field in ::fields track by $index" compile="::field.getTemplateValueWithLabel(entry)">
```
RichardBradley: The "form-field" and "form-group" classes here are redundant, since "ma-field" inside this div also creates a div with those classes. Having them repeated here interferes with hiding rows based on entity values.

doc/Relationships.md
```diff
@@ -223,6 +223,11 @@ When displaying a post:
 
 ![embedded_list in editionView](images/embedded_list_in_editionView.png)
 
+Caution: If you want to use [`Field.template()`](./Theming.md#customizing-the-template-for-a-given-field)
```
jpetitcolas: This paragraph is not clear to me. What do you mean? :)

RichardBradley: The docs for the "Field.template()" config in [Theming.md](https://github.com/marmelab/ng-admin/blob/faec147c4d7fcc06c0717cf4e357eeeb823e4f75/doc/Theming.md#customizing-the-template-for-a-given-field) suggest using it with an "<ma-field>" tag with 6 "magic" attrs that must be quoted verbatim, i.e.

    .template('<ma-field ng-if="entry.values.category" field="::field" value="entry.values[field.name()]" entry="entry" entity="::entity" form="formController.form" datastore="::formController.dataStore"></ma-field>', true)

Those 6 attrs expose the implementation details of ng-admin and they must be gotten exactly right for things to work. To make this even harder for users, the magic attrs that you must quote are different for "regular" fields to "embedded" fields.

Actually fixing this is out of scope here, but I thought users might benefit from a warning about this, in case they try to use "template()" on an embedded field and can't figure out why it didn't work.

jpetitcolas: Right: we should think to fix this black magic. :+1: 

Can you add your explanation directly into the README?

RichardBradley: i'd be happy to, but I'm not quite clear where you want this explanation. or what text you want in addition to what I've already put, sorry. Perhaps you should add an explanation that makes sense to you directly to the code? Or let me know in more detail what changes you had in mind, and I'll make them.

(Also perhaps this issue about template() would be better addressed on its own Issue thread, as it's separate from the problem in #1158)

RichardBradley: (bump)

RichardBradley:
Are you waiting for something from me on this?

From my perspective, this is ready to merge. It doesn't fix the "black magic", but it's a step in the right direction.

I haven't added any unit tests, as I feel like #1158 will be really difficult to demo in a unit test. Is that a blocker for merging this?

Kmaschta:
@RichardBradley I'll take a look at both your PR and issue this afternoon.

Kmaschta:
Ok, it's a tough one and I'm not sure that your fix is optimal.
I'll dig further on this later, but I can confirme that the `embedded_list` is the only field that doesn't support `.template(..., true)` for now.

RichardBradley:
> Ok, it's a tough one and I'm not sure that your fix is optimal.

Perfect is the enemy of good :-)

It's not optimal, but it's better than what's currently there and it doesn't make anything worse. I have been using this fix in production since shortly before I posted it here. I would suggest merging it while you work on an optimal fix in parallel.

It's up to you, of course. I can continue to use a forked version with this fix applied if you don't want it in here.

Thanks for your time and hard work on `ng-admin`.

RichardBradley:
(From my point of view, this is ready to merge)

PR #1172, opened 2016-08-04
Merge this build to fix the error of notification

fakharkhan:


fzaninotto:
I've updated the build files on my side. Can you confirm it fixes the issue?

PR #1189, opened 2016-09-12
[RFR] Fix non-dev dependencies

jpetitcolas:
As it breaks other projects because of missing dependencies (for instance `admin-config` one). 

Still in WIP as I am testing it on a project rewrite.


package.json
```diff
@@ -11,46 +11,56 @@
...
-  "devDependencies": {
+    "angular": "~1.4.8",
     "admin-config": "0.12.1",
-    "angular-mocks": "~1.4.8",
     "angular-numeraljs": "^1.1.6",
     "angular-sanitize": "^1.3.15",
     "angular-translate": "^2.11.0",
     "angular-ui-bootstrap": "~1.2.1",
-    "angular-ui-codemirror": "^0.3.0",
+    "angular-ui-codemirror": "jpetitcolas/ui-codemirror#di",
```
jpetitcolas: To fix some strict DI errors, preventing JSON field to be colored.

jpetitcolas:
Should be fine now. Can someone give it a try using this branch?

Phocea:
Happy to report that I did not face any more issues using this branch!

Since I am also using webpack and babel. I also managed to make this work without the transpiled libs, by using an alias and including the source of ng-admin.
If you think this is worth mentioning I will gladly add this to the documentation once this branch is merged.
Thanks @jpetitcolas !

jpetitcolas:
@Phocea: Glad to read it! :)

I voluntarily did not mention it. Indeed, if you use Babel, you can either use `src/` (ES6 files) or `lib/` (ES5 files). But if you don't, you can still use ES5. And as this is the default location of the lib (at least once published on `npm`), it should be fine. 

@fzaninotto: switching it to RFR. We should release an `alpha5.1` ASAP if everything is fine for you.

webpack.config.js
```diff
@@ -1,4 +1,5 @@
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
+var path = require('path');
```
fzaninotto: not used

PR #1197, opened 2016-09-19
[RFR] Fix error messages on failed deletes, updates

jpetitcolas:


jpetitcolas:
Closes #1129 and #1130.

RichardBradley:
(It doesn't close all of #1129, just makes it better. See comments on #1130)

Phocea:
@jpetitcolas do you have an ETA on this merge by any chance ?

jpetitcolas:
@Phocea: a few seconds ago. :)

Phocea:
works a treat, much better.
Do you think this one could also be merged https://github.com/marmelab/ng-admin/pull/1223 so we also get the error message displayed correctly when happening on state change?

Will open another PR later with a way to configure the redirect depending on  the http code error

PR #1208, opened 2016-09-22
Use lodash instead of underscore to avoid conflict with Restangular

Phocea:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "restangular": "~1.5.2",
     "textangular": "1.4.3",
     "ui-select": "0.18.1",
-    "lodash": ">=1.3.0"
+    "lodash.debounce": "4.0.8"
   },
   "devDependencies": {
     "angular-mocks": "~1.4.8",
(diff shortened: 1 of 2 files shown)


jpetitcolas:
Apparently, we are using underscore only for debouncing:

```
jpetitcolas@jpetitcolas-laptop ~/dev/ng-admin $ git grep "_\."
src/javascripts/ng-admin/Crud/list/ListLayoutController.js:            _.debounce((newValues, oldValues) => {
src/javascripts/ng-admin/Main/view/404.html:      |,4-  ) )-,_..;\ (  `'-'
src/javascripts/test/app-test.js:for (file in window.__karma__.files) {
src/javascripts/test/app-test.js:    if (window.__karma__.files.hasOwnProperty(file)) {
```

Perhaps can we just embed `loadash.debounce` lib in this case?

fzaninotto:
You need to import lodash in src/javascripts/ng-admin/Crud/list/ListLayoutController.js. 

I can't reopen this PR, can you make a new one?

PR #1210, opened 2016-09-22
add embedded_list formatter

JacerOmri:
Add basic formatter for embedded_list fields
It would map the array items to the value of the first target field configured

### DIFF
--- a/src/javascripts/ng-admin/Crud/misc/EntryFormatter.js
+++ b/src/javascripts/ng-admin/Crud/misc/EntryFormatter.js
@@ -75,6 +75,13 @@ export default class EntryFormatter {
                         value: entry.listValues[field.name()].join(', ')
                     };
                 };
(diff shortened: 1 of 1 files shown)


src/javascripts/ng-admin/Crud/misc/EntryFormatter.js
```diff
@@ -75,6 +75,13 @@ export default class EntryFormatter {
                         value: entry.listValues[field.name()].join(', ')
                     };
                 };
+            case 'embedded_list':
+                return function (entry) {
+                    return {
+                        name: label,
+                        value: entry.values[field.name()].map((obj => obj[field.targetFields()[0].name()]))
```
jpetitcolas: Perhaps would it be more coherent to use the same value as for the `reference_many`?

```
value: entry.listValues[field.name()].join(', ')
```

PR #1215, opened 2016-09-28
[RFR] Unescaped translation

### LINKED ISSUE #1214: Angular Translate automatically HTML escape all string fields
### Description

Since `angular-translate` is installed on ng-admin, all string fields are HTML escaped.

It is caused by the line `$translateProvider.useSanitizeValueStrategy('sanitize');` in [config/translate.js](https://github.com/marmelab/ng-admin/blob/f35c2cede1d9219b8aade65440bf25218df1d4e9/src/javascripts/ng-admin/Main/config/translate.js#L45).

The [angular-translate's documentation](https://angular-translate.github.io/docs/#/guide/19_security) recommand to set the sanitize strategy to `strategy` but we want to be able to switch it off.

### Steps to Reproduce

You have an API endpoint that serve this entity:

```json
{
    "id": 42,
    "name": "TP & SP"
}
```

And a ng-admin config as following:

```js
entity.listView().fields([
    nga.field('id', 'number'),
    nga.field('name', 'string'),
]);
```

**Expected behavior:**

field | value
---- | ----
id | `42`
name | `TP & SP`

**Actual behavior:**

field | value
---- | ----
id | `42`
name | `TP &amp; SP`

Kmaschta:
Disable default HTML escape from angular-translate

Fix #1214


Kmaschta:
@jpetitcolas Done, I haved to do it with a E2E test

Issue #1220, opened 2016-10-03
Update PlunkR with last version


jpetitcolas:
We should update/fork [current PlunkR](http://plnkr.co/edit/qrw9Y8rfUUmHM4rOJo1q?p=preview) to use last version of ng-admin, using following URL: 

* https://unpkg.com/ng-admin/build/ng-admin.min.js
* https://unpkg.com/ng-admin/build/ng-admin.min.css

They point to last built version published on npm.

PR #1221, opened 2016-10-03
Update issue template with proper plunker

Phocea:
work with ng admin > 1.0.0-alpha7

Fix  https://github.com/marmelab/ng-admin/issues/1220


fzaninotto:
Awesome, thanks!

PR #1223, opened 2016-10-04
Fix state change error messaging

Phocea:
error object is composed of:  data, status, statusText and config ... there is no message attributes


Phocea:
@jpetitcolas This at least would handle all kind of error to be displayed properly. However how could I grab message customization like it is done via [app.errorMessage](https://github.com/marmelab/ng-admin/blob/master/doc/Theming.md#customizing-error-messages).

May also be useful to customise the routing depending on the HTTP code. Is this something that could be added inside the global config object in a map of some sort. Maybe something like .

app.setErrorRouting('htttpCode', 'route')

Phocea:
Closing this for now .Fix will be part of a bigger change

PR #1226, opened 2016-10-11
[documentation] changed the missing ui-view name (else the sample is not working, and a…

dealboy:
fixed the missing ui-view name (else the sample is not working, and an absolute beginner gets lost :(  )


Kmaschta:
Thank you!

PR #1234, opened 2016-10-15
Edit Api-mapping nested relationships documentation.

PrintHelloWorld:
- Changed filter to _filters to be more consistent with current version of ng-admin


jpetitcolas:
Thanks! :)

Issue #1238, opened 2016-10-24
Release SASS files to a separate subdirectory


SeMeKh:
Currently, installing `ng-admin@1.0.0-beta2` using `npm` results in a `lib` directory structure like this:

```
lib
├── javascripts
│   ├── ng-admin
│   ├── ng-admin.js
│   ├── ng-admin.js.map
│   ├── test
│   ├── vendors.js
│   └── vendors.js.map
├── ng-admin.scss
└── sb-admin.scss
```

Based on notes from [Better Integration with Module Bundlers](https://github.com/marmelab/ng-admin/blob/master/UPGRADE-1.0.md) I assume both `.scss` files should be moved to a `sass` sub-directory.

PR #1248, opened 2016-11-18
[RFR] Fix SCSS reference in the 1.0 changelog

Kmaschta:
:warning: We need to choose either
- Merge this PR
- Address the issue #1238


fzaninotto:
No, we'll take the other way around (changing the path where the scss files are copied)

PR #1252, opened 2016-11-22
Use lodash instead of underscore

Phocea:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "restangular": "~1.5.2",
     "textangular": "1.4.3",
     "ui-select": "0.18.1",
-    "underscore": "^1.8.3"
+    "lodash": "4.17.2"
   },
   "devDependencies": {
     "angular-mocks": "~1.4.8",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "restangular": "~1.5.2",
     "textangular": "1.4.3",
     "ui-select": "0.18.1",
-    "lodash": "4.17.2"
+    "lodash.debounce": "4.0.8"
   },
   "devDependencies": {
     "angular-mocks": "~1.4.8",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "restangular": "~1.5.2",
     "textangular": "1.4.3",
     "ui-select": "0.18.1",
-    "underscore": "^1.8.3"
+    "lodash": "4.17.2"
```
fzaninotto: we just need `lodash.debounce`, please don't add the entire `lodash` lib as dep.

src/javascripts/vendors.js
```diff
@@ -12,4 +12,4 @@ require('angular-numeraljs');
 require('angular-ui-bootstrap/dist/ui-bootstrap-tpls');
 require('ng-file-upload');
 
-global._ = require('underscore');
+global._ = require('lodash');
```
fzaninotto: no, this won't work anymore now that we don't have lodash as dependency

Phocea: I know, but adding var debounce = require('lodash.debounce'); into ListLayoutController will right ?

Phocea: ..actually just tested, and if I reduce this to lodash.debounce, it is ok for the core code of ng-admin but it then make all Restangular interceptor code to fail as specified in the doc...
So not sure how this should be handled

fzaninotto:
As seen with @Phocea, `lodash` is used by `Restangular` anyway. Adding `lodash.debounce` is then counterproductive, since we can't reuse the full `lodash`. So it's better to depend on `lodash` than `lodash.debounce`.

Phocea:
Rebased and replaced by https://github.com/marmelab/ng-admin/pull/1268

PR #1257, opened 2016-11-30
[WIP] Better http error handling

Kmaschta:
@Phocea Following https://github.com/marmelab/ng-admin/pull/1253


Kmaschta:
@jpetitcolas There is still an issue with protractor or SauceLabs. Can you take a look?

fzaninotto:
@Phocea the build fails, but it seems to be because of a problem in your code rather than a build artefact. Can you take a look?

jpetitcolas:
Tests should be fixed now, **but** we still need to check new feature. I quickly copy/pasted README sample, but it doesn't boot the application anymore. Back to WIP, but we should not be too far. :)

src/javascripts/ng-admin/Main/component/provider/HttpErrorService.js
```diff
@@ -1,38 +1,36 @@
...
-        displayError: text => notification.log(text, { addnCls: 'humane-flatty-error' }),
-    }
-}
+    displayError: text => notification.log(text, { addnCls: 'humane-flatty-error' }),
+});
 
 httpErrorService.$inject = ['$state', '$translate', 'notification'];
 
-export default httpErrorService;
+export default { $get: httpErrorService };
```
jpetitcolas: @Phocea: this is a provider, hence we need a `$get` method. :)

Phocea:
Replaced by https://github.com/marmelab/ng-admin/pull/1257

PR #1266, opened 2016-12-07
[WIP] Npm postinstall for linux / windows

Phocea:
Follow up on [E2E fix commit](https://github.com/marmelab/ng-admin/commit/9a845e41eb0a63604d4483fc410fdb6093df139d), so the npm postinstall phase also work on Windows node installations

PR #1267, opened 2016-12-08
Npm post install for windows

Phocea:
Follow up on E2E fix commit, so the npm postinstall phase also work on Windows node installations
Rebase of https://github.com/marmelab/ng-admin/pull/1266

### DIFF
--- a/_postinstall.js
+++ b/_postinstall.js
@@ -1,0 +1,42 @@
+/*
+ * _postinstall.js is a script that runs automatically after the `npm install`
+ */
+// Get platform from node
+var os = require('os');
+var platform = os.platform();
+
+if (platform === 'darwin' || platform == 'linux') {
+  // Call child process and execute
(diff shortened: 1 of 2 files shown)


Phocea:
@Kmaschta .. Rebasing done the trick. So this is the way to go until we figure out the problem with the tests....

Kmaschta:
Hmmm.. Does it work with `node ./node_modules/protractor/bin/webdriver-manager update --versions.chrome=2.24` in the package.json?

Phocea:
Nope, see https://github.com/marmelab/ng-admin/commit/9a845e41eb0a63604d4483fc410fdb6093df139d#commitcomment-20044668

Kmaschta:
There is still an issue with the postinstall, can you figure out why?

image

Phocea:
Looks more like a protractor isseu you are getting there. I restarted Travis build and its fine in their VM

Issue #1209, opened 2016-09-22
Using underscore in vendor.js seems to conflict with Restangular


Phocea:
### Description

Using the 1.0.0-alpha7 version build I am getting the error _.includes is not a function when acessing any of my entity list views

Following several threads on Restangular Git (one of them being mgonto/restangular#1225). I found this is a know issue with Restangular and various changes in the underscore and lodash librabr
Restangular 1.5.2 is now using "lodash": ">=1.3.0" which fixes the issue.
However using underscore for the _ global value offuscate it to Restangular.

This issue only occur since commit cc21984e7827389d6de06ed86aee8f6664c6ba63 (webpack dependencies commit).
It occurs if I use the already transpiled libs, or transpile them myself using webpack. Before this, using the minified build if ng-admin-only did not cause any problem.

Changing the following line in vendor.js fixes the problem but your CI does not like it...
from:
`global._ = require('underscore');`
to:
`global._ = require('lodash');`

The error stack encountered is (ml-sp-opui.js.js is my webpacked app):

`angular.js:12798 TypeError: _.includes is not a function at Object.t.isSafe (restangular.min.js:6) at Object.O (restangular.min.js:6) at Object.A (restangular.min.js:6) at e.value (RestWrapper.js:19) at t.value (ReadQueries.js:19) at Object.e.state.resolve.rawEntry (routing.js:166) at Object.invoke (angular.js:4570) at c (angular-ui-router.js:475) at angular-ui-router.js:467 at processQueue (angular.js:15112)`

jpetitcolas:
@Phocea: I can't reproduce the issue on the example app (using `make run`), even after a brand new `npm install`. Can you pinpoint the issue more precisely?

Phocea:
I will have another go and also check if the issue occur with the build under Linux.
At the moment I have got around the problem by automatically patching the vendor.js during my Maven Windows build, prior to do the webpack and test phase.
In any case, it seems that lodash is a better choice for the future than underscore (which it includes), but this is another discussion :)

Phocea:
@jpetitcolas I still have the issue both on windows and linux build and really cant see what could cause this side effect.
Best may be too close this PR since I have a valid workaround, and maybe create an enhancement with low priority to move to lodash in the future.
You could also ask @fzaninotto if he spotted any issues going to lodash rather than underscore when doing this task: https://github.com/marmelab/ng-admin/pull/845

beatelite:
[...]

PR #1268, opened 2016-12-08
Use lodash instead of underscore

Phocea:

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,7 @@
     "rangy": "^1.3.0",
     "restangular": "~1.5.2",
     "textangular": "1.4.3",
-    "ui-select": "0.18.1",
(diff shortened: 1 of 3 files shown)


Phocea:
Rebase and replace https://github.com/marmelab/ng-admin/pull/1252

Phocea:
Fix https://github.com/marmelab/ng-admin/issues/1209

Kmaschta:
If you already have lodash, why do you need to add `lodash.debounce` in dependencies?

And don't you need `lodash` in dependencies instead of devDeps?

Phocea:
lodash doesnt include lodash.debounce by default.
lodash is needed by Restangular so not only in dev.

Issue #1263, opened 2016-12-05
no comments shown with example code from "getting started" tutorial


chkuendig:
### Description
I just tried to go through the Getting Started Tutorial, but for some reason i can't get the referenced_lists thing to work: https://ng-admin-book.marmelab.com/doc/Getting-started.html
 

### Steps to Reproduce
http://plnkr.co/edit/uq1ZFiXUPuqg1o749k6O?p=preview (copy-pasted admin.js from the bottom of the Getting Started page)
click on one of the posts, you'll see an empty list with 5 rows

**Expected behavior:** 
![Getting Started Example](https://ng-admin-book.marmelab.com/doc/images/post_show_view_with_related_comments.png) 

**Actual behavior:** 
click on one of the posts, you'll see an empty list with 5 rows
<img width="527" alt="screen shot 2016-12-05 at 18 49 59" src="https://cloud.githubusercontent.com/assets/748698/20895879/9ad9c318-bb1b-11e6-9789-4ca7d6a36376.png">

Phocea:
The error in the plunkr shows

```
ng-admin.min.js:7 Error: Unable to find entity "comments"
    at e.value (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:26:25661)
    at pre (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:45:9654)
    at https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:7:9042
    at ae (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:7:9099)
    at g (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:7:2673)
    at s (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:6:31254)
    at https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:6:30901
    at link (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:45:3072)
    at https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:7:9042
    at ae (https://unpkg.com/ng-admin@1.0.0-beta3/build/ng-admin.min.js:7:9099) <ma-referenced-list-column field="::field" datastore="::datastore" class="ng-scope ng-isolate-scope">
```
[...]

PR #1270, opened 2016-12-08
Fix Getting started code and documentation

Phocea:
Remove the note abut nonexistent entity being created automatically
Fix the getting started code


Phocea:
Fix https://github.com/marmelab/ng-admin/issues/1263

Issue #1187, opened 2016-09-09
Replace name  property by label in "Add New" button in embedded_list fields


tgeorges:
### Scope

When using embedded list fields, we cannot change the add XXX button label and it always displays the programmatic name. A good default behavior can be to show the label : 

`<span translate="ADD_NEW" translate-values="{ name: field().name() }"></span>`

Replaced by 

`<span translate="ADD_NEW" translate-values="{ name: field().label() }"></span>`

### Prerequisites
ng-admin 1.0.0-alpha5

### Description

All is in the scope

### Steps to Reproduce

Add an embedded_list field to an entity
Set the label

**Expected behavior:** [What you expected to happen]
The add button displays the label
**Actual behavior:** [What actually happened]
The add button displays the name

jpetitcolas:
Duplicate of https://github.com/marmelab/ng-admin/issues/1114

PR #1281, opened 2016-12-21
[RFR] Use field label for embedded list add button

### LINKED ISSUE #1114: Embedded List field add button text
Button text currently uses raw field name (with underscores and all).
Instead it should use label and singularize it.

jpetitcolas:
Fixes #1114, #1187.


jpetitcolas:
For a better looking button. I prefer "Add a link"  to Add a Link". But I'm not a native English speaker. Perhaps standard form would be better?

Phocea:
Isn't it going to be "add a link" if all lowercase though ?
Just let whatever was set in the label decide no ?

PR #1289, opened 2017-01-07
fix sass filepath in example

collinforrester:


Kmaschta:
Thank you for taking the time to do a PR.

The SASS file issue (#1238) was fixed by my PR #1261 and will be deployed with the next major version ASAP.

The postinstall script doesn't work on your machine?
Please open an issue or make an isolated PR.

Idem for the `angular-numeraljs` update. What's the issue?

Finally, you cannot install the project directly from GitHub without the build phase which is ran before the NPM publication.

I close the PR for the reasons listed above, it's better to make multiple tiny PR.
Thanks again!

collinforrester:
Sorry for the multiple changes! I forgot I submitted the PR on master and I kept making changes to it.

angular-numeraljs is depending on a version of numeral that isn't importing correctly with webpack.  The issue is described [here](https://github.com/adamwdraper/Numeral-js/issues/403) and fixed in the most recent version of numeraljs.

I had forked this project and angular-numeraljs in an attempt to update all the dependencies and test a build like that.

Thanks for responding.

PR #1291, opened 2017-01-09
[RFR] Check for fit, fdescribe or .only in tests

Kmaschta:
In order to avoid `fit`, `fdescribe` or `.only` on master.


Kmaschta:
And it works: https://travis-ci.org/marmelab/ng-admin/builds/190247672

image

jpetitcolas:
image

It would worth to use it everywhere! :+1:

PR #1300, opened 2017-01-19
[RFR] Handle CRUD form error through http error service

Phocea:
This PR is aim to make use of the new HttpErrorService to display CRUD form http error responses.
For backward compatibility, the code still pick up customization done via [entity.errorMessage method](https://github.com/marmelab/ng-admin/blob/Handle-Crud-Form-Error-Through-HttpErrorService/doc/Theming.md#customizing-error-messages). However this step is now redudant since it can be done inside an decorator

## TODO
- [x] More specific unit tests?
- [x] Discuss with the team if this is the correct implementation.
- [x] Update doc

### DIFF
--- a/doc/Theming.md
+++ b/doc/Theming.md
@@ -181,6 +181,8 @@ entity.errorMessage(function (response) {
     return 'Global error: ' + response.status + '(' + response.data + ')';
 });
 ```
+If you do not customize the error message using that method, it will be globally handled by the [Http Error Service](../src/javascripts/ng-admin/Main/component/provider/HttpErrorService.js).
+
 ## Customizing HTTP Error Messages
 
 If you want to override, patch or extend the way HTTP errors are handled by the ng-admin [Http Error Service](../src/javascripts/ng-admin/Main/component/provider/HttpErrorService.js) then you may use an [Angular decorator](https://docs.angularjs.org/guide/decorators), as per the below example:
--- a/doc/Theming.md
+++ b/doc/Theming.md
@@ -215,3 +215,4 @@ Bind the decorator to your application:
 ```js
 myApp.decorator('HttpErrorService',require('HttpErrorDecorator'));
 ```
+The decorator can also be used to globally customize CRUD error messages
--- a/src/javascripts/ng-admin/Crud/delete/DeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/DeleteController.js
@@ -1,5 +1,5 @@
 export default class DeleteController {
-    constructor($scope, $window, $state, $q, $translate, WriteQueries, Configuration, progression, notification, params, view, entry) {
+    constructor($scope, $window, $state, $q, $translate, WriteQueries, Configuration, progression, notification, params, view, entry, HttpErrorService) {
         this.$scope = $scope;
         this.$window = $window;
         this.$state = $state;
@@ -17,6 +17,7 @@ export default class DeleteController {
         this.notification = notification;
         this.$scope.entry = entry;
         this.$scope.view = view;
+        this.HttpErrorService = HttpErrorService;
 
         $scope.$on('$destroy', this.destroy.bind(this));
 
@@ -45,14 +46,14 @@ export default class DeleteController {
             .then(() => $translate('DELETE_SUCCESS'))
             .then(text => notification.log(text, { addnCls: 'humane-flatty-success' }))
             .catch(error => {
-                const errorMessage = this.config.getErrorMessageFor(this.view, error) || 'ERROR_MESSAGE';
+                const errorMessage = this.HttpErrorService.handleError(this.config.getErrorMessageFor(this.view, error) || 'ERROR_MESSAGE');
                 progression.done();
-                $translate(errorMessage, {
-                    status: error && error.status,
-                    details: error && error.data && typeof error.data === 'object' ? JSON.stringify(error.data) : {}
-                })
-                    .catch(angular.identity) // See https://github.com/angular-translate/angular-translate/issues/1516
-                    .then(text => notification.log(text, { addnCls: 'humane-flatty-error' }));
+                //$translate(errorMessage, {
+                //    status: error && error.status,
+                //    details: error && error.data && typeof error.data === 'object' ? JSON.stringify(error.data) : {}
+                //})
+                //    .catch(angular.identity) // See https://github.com/angular-translate/angular-translate/issues/1516
+                //   .then(text => notification.log(text, { addnCls: 'humane-flatty-error' }));
             });
     }
 
@@ -70,7 +71,8 @@ export default class DeleteController {
         this.entity = undefined;
         this.progression = undefined;
         this.notification = undefined;
+        this.HttpErrorService = undefined;
     }
 }
 
-DeleteController.$inject = ['$scope', '$window', '$state', '$q', '$translate', 'WriteQueries', 'NgAdminConfiguration', 'progression', 'notification', 'params', 'view', 'entry'];
+DeleteController.$inject = ['$scope', '$window', '$state', '$q', '$translate', 'WriteQueries', 'NgAdminConfiguration', 'progression', 'notification', 'params', 'view', 'entry', 'HttpErrorService'];
--- a/src/javascripts/ng-admin/Crud/delete/DeleteController.js
+++ b/src/javascripts/ng-admin/Crud/delete/DeleteController.js
@@ -46,14 +46,12 @@ export default class DeleteController {
             .then(() => $translate('DELETE_SUCCESS'))
             .then(text => notification.log(text, { addnCls: 'humane-flatty-success' }))
             .catch(error => {
-                const errorMessage = this.HttpErrorService.handleError(this.config.getErrorMessageFor(this.view, error) || 'ERROR_MESSAGE');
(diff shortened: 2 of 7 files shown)


Phocea:
@Kmaschta @jpetitcolas @fzaninotto 
Your thoughts on this PR.
Would it also be acceptable to start marking entity.getErrorMessage() as deprecated to only keep ony way of customizing error messages using a decorator (and simplify ng-admin). Or we can keep and document both ways.

Kmaschta:
I think that [this change](https://github.com/marmelab/ng-admin/pull/1300/files#diff-eaebe4c626350de4bb3a4ce390c6b416R3) should be done with the 1.0.

Phocea:
I agree it would be neater to submit this in the 1.0 together with the HttpErrorService. Can we postpone while we get more feedback on this?

Kmaschta:
Have you something to add to this PR or can we merge it?

Phocea:
I need to update the documentation. Do we keep the dual way of customizing the error message (decorator or entity.errorMessage(), or do we mark the later as deprecated?

Also, since the [HttpErrorService ](https://github.com/marmelab/ng-admin/blob/master/src/javascripts/ng-admin/Main/component/provider/HttpErrorService.js)is used for both CRUD and stateChangeError, then I think it would make more sense for the defaultError to only gice the STATE_CHANGE_ERROR of the toState and such parameters exists, otherwise just display a GLOBAL_ERROR of some sort

Phocea:
@Kmaschta, @jpetitcolas Any of you guys have a chance to review and  merge those changes ?

Phocea:
@Kmaschta I agree, but the more we wait the more troublesome that BC could be. Right now I doubt if its going to force people to make code changes...except from me :)

doc/Theming.md
```diff
@@ -215,3 +217,4 @@ Bind the decorator to your application:
 ```js
 myApp.decorator('HttpErrorService',require('HttpErrorDecorator'));
 ```
+The decorator can also be used to globally customize CRUD error messages
```
jpetitcolas: s/messages/messages./

src/javascripts/ng-admin/Crud/delete/DeleteController.js
```diff
@@ -45,14 +46,12 @@ export default class DeleteController {
             .then(() => $translate('DELETE_SUCCESS'))
             .then(text => notification.log(text, { addnCls: 'humane-flatty-success' }))
             .catch(error => {
-                const errorMessage = this.config.getErrorMessageFor(this.view, error) || 'ERROR_MESSAGE';
+                // this is kept for backward compatibility with the entity.errorMessage() method
```
jpetitcolas: Can you add a `@deprecated` tag? This way, when we'll ship a `2.0` version, we could `grep` all these tags to make some cleanup. :)

src/javascripts/ng-admin/Main/config/translate.js
```diff
@@ -39,6 +39,7 @@ export default function translate($translateProvider) {
       'DETAIL': 'Detail',
       'STATE_CHANGE_ERROR': 'State change error: {{ message }}',
       'STATE_FORBIDDEN_ERROR': 'A server 403 error occured: {{ message }}',
+      'GLOBAL_ERROR': 'Global error: {{ message }}',
```
jpetitcolas: I would instead just specify `Error:`, as global is not really explicit for an end-user. :)

src/javascripts/ng-admin/Crud/form/FormController.js
```diff
@@ -71,20 +72,18 @@ export default class FormController {
             .then(() => $translate('CREATION_SUCCESS'))
             .then(text => notification.log(text, { addnCls: 'humane-flatty-success' }))
             .catch(error => {
-                const errorMessage = this.config.getErrorMessageFor(this.view, error) || 'ERROR_MESSAGE';
+                // this is kept for backward compatibility with the entity.errorMessage() method 
+                const msg = this.config.getErrorMessageFor(this.view, error) || 'ERROR_MESSAGE';
+                error.data.message = msg;
+                
+                const errorMessage = this.HttpErrorService.handleError(error);
```
jpetitcolas: Can you add some unit tests to check that the `HttpErrorService` is correctly called?

jpetitcolas:
I agree with @Kmaschta: even if we didn't explicitly encourage to use `HttpErrorHandler` directly, some people may use them, through a decorated handler for instance. To respect semantic versioning, I would recommend to keep compatibility as currently, and to update it when switching to a `2.0` version. 

Unless we release discretly (without a blog post) a new major version after each BC change? @fzaninotto, what do you think about it?

Kmaschta:
As discussed with @fzaninotto & @jpetitcolas, we'll most likely prepare a `2.0`.
Meanwhile, we can merge all the BC features on a new branch: `next`.

jpetitcolas:
Oops, my bad. Merged too fast. Read "we can merge" from @Kmaschta. @Phocea, can you fix your review in a new PR? :)

src/javascripts/ng-admin/Crud/form/FormController.js
Phocea: So thats an entirely new test on FormController right? Could not see any existing ones?

Phocea: @jpetitcolas, I tried to write that damn test case but its proving to be quiet a pain. The reason is that  FormController is rather a big class, and I want to only test a very small part of it.
Since I dont have big experience with Karma, I may going at this the wrong way. I have mocked quiet a few services, and tried to use .and.callThrough(), but when invoking writeQueries it is all falling apart :(
Would you mind having a quick look to see where I am going wrong ? Thanks !


```
const  FormController = require('../../../js/core/controllers/FormController.js');

describe('To Test FormController', function(){
    let mockHttpErrorService;
    let mockFormController;
    let view;
    let $state;
    let $injector;
    let $translate;
    let previousState;
    let WriteQueries;
    let Configuration;
    let progression;
    let notification;
    let dataStore;
    let $scope;
    let $event;
    let $rootScope;

    angular.module('testFormController', [])
    .service('Configuration',() => ({
    }))
    .service('view', function() {
        return {
            title: jasmine.createSpy('view.title'),
            description: jasmine.createSpy('view.description'),
            actions: jasmine.createSpy('view.actions'),
            fields: jasmine.createSpy('view.fields'),
            getEntity: function(){
                return {
                    uniqueId:"",
                    identifier:function(){
                        return{
                            name:function(){return "val"}//jasmine.createSpy('view.getEntity.identifier.name')
                        }
                     }
                }
                } 
        };
    })
    .service('$state', function() {
        return {
        };
    })
    .service('$event', function() {
        return {
            preventDefault:jasmine.createSpy('$event.preventDefault')
        };
    })
    .service('dataStore', function() {
        return {
            getFirstEntry:function(id){
                return{
                    values:{}
                }
            }
        };
    })
    .service('$translate', function() {
        return {
        };
    })
    .service('previousState', function() {
        return {
        };
    })
    .service('WriteQueries', function() {
        return {
            
            createOne:jasmine.createSpy('WriteQueries.createOne')
       };
    })
    .service('progression', function() {
        return {
       };
    })
    .service('notification', function() {
        return {
        };
    })
    .service('HttpErrorService', function() {
        return {
            handleError: jasmine.createSpy('HttpErrorService.handleError')
        };
    })
  .controller('FormController', ['$scope','$state','$injector', '$translate', 'previousState', 'WriteQueries', 'Configuration', 'progression', 'notification','view','dataStore','HttpErrorService',FormController]);

    beforeEach(angular.mock.module('testFormController'));

    beforeEach(inject(function(_$event_,_HttpErrorService_, $controller, _view_,_$state_, $injector, _$translate_, _previousState_, _WriteQueries_, _Configuration_, _progression_, _notification_,_dataStore_) {
        Configuration= _Configuration_;
       $rootScope = $injector.get('$rootScope');
        $scope = $rootScope.$new();
        view = _view_;
        $state= _$state_;
        dataStore=_dataStore_;
        $translate= _$translate_;
        previousState= _previousState_;
        WriteQueries= _WriteQueries_;
        progression= _progression_;
        notification= _notification_;
        $event = _$event_;
        mockHttpErrorService = _HttpErrorService_;
        mockFormController = $controller('FormController', {$scope:$scope,$state:$state, $injector:"", $translate:$translate, previousState:previousState, WriteQueries:WriteQueries, Configuration:function conf(){}, progression:progression, notification:notification,view:view, dataStore:dataStore,HttpErrorService:mockHttpErrorService});
    }));

    it('should be able to call HttpErrorService', () => {
        var error = {status:403}
        spyOn(mockFormController, 'validateEntry')
                spyOn(mockFormController, 'WriteQueries')

        mockFormController.submitCreation($event);
        try {
            expect(WriteQueries.createOne().toThrow(new Error("403")));
       } catch (e) {
       }
        
    });

});

```

jromero-onclick:
Hello at this moment should I have to decorate HttpErrorService for accomodate form errors with my api or not? I did not found the steps on the doc

Phocea:
Hello, Documentation is here: https://github.com/marmelab/ng-admin/blob/master/doc/Theming.md#customizing-http-error-messages

Until those changes are merged on main you will only be able to accomodated for stateChange errors.
For API CRUD errors you have to wait for this fix. Should not be too long now

PR #1307, opened 2017-02-13
Makefile: Fix generating source-map with webpack

beevelop:
ATM there are no sourcemaps available, what is quite a pain to debug. It seems as if they were intended to exist in the first place.

The command in the Makefile is slightly different to really get sourcemaps for the `ng-admin.min.js` and `ng-admin.min.css` files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ transpile:
 
 build:
 	make transpile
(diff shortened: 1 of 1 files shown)


Phocea:
@Kmaschta can we merge this. Creating the sourcemap can take a bit of time but its indeed useful for debugging

Kmaschta:
I merge it, but I would love a double check from @fzaninotto

PR #1312, opened 2017-02-27
Update Field.md

escapedcat:
Fix anchor jump links.
If `-` has a special function let me know. Here the jump-liks do not work:
https://ng-admin-book.marmelab.com/doc/reference/Field.html#-referenced-list-field-type


Kmaschta:
Thanks!

Issue #1305, opened 2017-02-08
Date field cannot be typed in


Joe-Edwards:
If you have a 'date' field, you can no longer type the date in it explicitly. Attempting to modify an existing date also just deletes it.

You can see the behaviour in the ng-admin demo - https://marmelab.com/ng-admin-demo/index.html#/customers/edit/485 - try changing the birthday without opening the picker.

I think this was broken by https://github.com/marmelab/ng-admin/issues/899 - it used to work in an older version of ng-admin.

The problem is because maDateField watches both value and rawValue:

```
            scope.$watch('rawValue', function(newValue) {
                scope.value = field.parse()(newValue);
            });

            scope.$watch('value', (newValue, oldValue) => {
                if (newValue === oldValue) {
                    return;
                }

                if (!newValue) {
                    scope.rawValue = null;
                    return;
                }

                scope.rawValue = scope.value instanceof Date ? scope.value : new Date(scope.value);
            });
```

The problem is, if at any point an invalid date string is in the box - suppose the string 'foo'. The following sequence happens:
- `$watch('rawValue'` sees `foo`. This is not a date object, so `parse` leaves it unchanged, and `value` is set to `foo`.
- `$watch('value'` sees `foo` and attempts to construct a new Date object using it. This ends up setting `rawValue` to an `Invalid Date` object.
- `$watch('rawValue'` sees `Invalid Date`. This is a date object, so `parse` tries to manipulate it, but fails and ends up setting the value to `null`. 
- `$watch('value'` sees `null` and correspondingly sets `rawValue` to `null`
- `$watch('rawValue'` sees `null` and correspondingly sets `value` to `null`
- `$watch('value'` sees `null`, this is the same as the last value, and the cycle ends!

The net result is `null` - which removes the text you tried to write from the box.

----

This is related to https://github.com/marmelab/ng-admin/issues/1271 which will be caused by a similar loop. However in that case, if you are in a timezone, the value/rawValue watches will loop, and because the 'parse' method applies a timezone correction the value never stabilises (resulting in the infinite loop).

mosasiru:
In many use cases, operators want to manage hour and minutes,  but it is not enabled now by the degration.
These cases are more popular than https://github.com/marmelab/ng-admin/issues/899, so  https://github.com/marmelab/ng-admin/issues/899 should be reverted once.

aikizoku:
Please  fix me (T_T)

matheo:
[...]

Issue #1311, opened 2017-02-25
reference_many: TypeError: Cannot read property 'identifierValue' of undefined


matheo:
### Scope

reference_may field throwing an exception following the documentation example. I'm using ng-admin 1.0.0

### Description

I'm setting up an **Array Of Foreign Keys** on my **Services** entity, to edit the related **Subservices**, but the edition form is throwing a weird error. I'm using the following form for my service:
```
  service.identifier(nga.field('_id'));
  service.editionView()
    .title('Edit a Service')
    .fields([
      nga.field('name'),
      nga.field('description', 'text'),
      nga.field('subservices', 'reference_many')
      .targetEntity(nga.entity('subservices'))
      .targetField(nga.field('name'))
      .singleApiCall(ids => ({'_id': ids })),
    ]);

```
My MongoDB **_id**s works ok in the panel in general, but this field type is throwing:
```
TypeError: Cannot read property 'identifierValue' of undefined
    at http://localhost:31227/ng-admin/ng-admin.min.js:47:831
    at Array.filter (native)
    at http://localhost:31227/ng-admin/ng-admin.min.js:47:804
    at Array.map (native)
    at i (http://localhost:31227/ng-admin/ng-admin.min.js:47:780)
    at Object.link (http://localhost:31227/ng-admin/ng-admin.min.js:47:1352)
    at http://localhost:31227/ng-admin/ng-admin.min.js:6:14391
    at http://localhost:31227/ng-admin/ng-admin.min.js:7:16776
    at be (http://localhost:31227/ng-admin/ng-admin.min.js:7:16833)
    at p (http://localhost:31227/ng-admin/ng-admin.min.js:7:9401) <ma-reference-many-field field="::field" value="value" datastore="::datastore" class="ng-isolate-scope">
```

I see that loading the form, the main entity (service) is `GET`ted but also a list of my subservices `GET /rest/subservices?limit=30&skip=0` (cache I think?), then there's a query of the related subservices (or multiple queries if `singleApiCall` is used, but then comes the exception handling the identifiers :(

### Steps to Reproduce

I tried to build my App to see if there was a better trace of the error with no joy, I'm thinking that it has something to do with the "_id" that is not being usedd from the configuration.

Any tips are really appreciated!

matheo:
[...]

Issue #786, opened 2015-11-13
embedded_list reference fields do not work


jamespsterling:
I have an edit view with an embedded list field that has sub-fields of reference type and they do not work properly. There is no call to the rest api asking for the list in this example.

```
        nga.field('editResultsId', 'embedded_list')
            .label('Edit Results')
            .targetFields([
                nga.field('fileSystemId', 'reference')
                    .label('File System')
                    .targetEntity(file)
                    .targetField(nga.field('name'))
                    .validation({ required: true }),
```

fzaninotto:
Hi, which version of ng-admin are you using? This has been added to 0.9.

jamespsterling:
I'm using 0.9

jamespsterling:
For clarification, it's the fileSystem reference child on this embedded list entity that does not populate. is this a bug?

fzaninotto:
Ah ok, now I get it: it's the `file` dropdown that is empty, is that it? Then it's just not implemented yet.

jamespsterling:
Yes, that's the issue.

clems71:
+1 would be very helpful indeed!

xteron:
Just stumbled over the same problem with reference_many inside a embedded_list.

NadavK:
I am not sure this is the same problem.
I am trying to reference a object with nested ids.
Maybe someone can help?


Object with nested references:
```
{"user":
  "id": 1,
  "orders": [
    {"order_id: 100},
    {"order_id": 101}
  ]
}
```
The code below returns the following error
> (state change error: e.targetField(...).name is not a function)

```
user.showView().fields([
    nga.field('orders.id', 'reference_many')
          .targetEntity(order)
          .targetField('id')
]);
```

(BTW, great project! Having an admin tool that works against the backend-logic instead of directly to the DB is truly cool).

mosasiru:
+1 seriously need!
is there any workaround?

collinforrester:
@NadavK I don't have a solution for the opened issue (which I also need solved).  But your error is because you don't have `targetField('id')` in a field definition: `targetField(nga.field('id'))` I think.

MaxTwentythree:
+1 need it too, would be really great :)

dowlingw:
Some notes from my duplicate over at #1145 (sorry!)

I can confirm this is still an issue in `1.0.0-alpha4`

1. `showView` does not perform any retrievals of `reference` objects for rendering
[...]

PR #1317, opened 2017-03-08
maDateField timezone adjustment. Fixes #1271 #1305

### LINKED ISSUE #1271: editionView() returns date decremented by 1 day
### Description

In my index.js:
```
project.editionView().fields([
         nga.field('devdate', 'date')
           .format('MM-dd-yyyy')
           .label('Development Date'),
    ]);
```
The day (dd) is decremented by 1 but the above works fine (displays the correct `devdate` response) in `project.creationView().fields`. 

### Steps to Reproduce
Plunkr: http://plnkr.co/edit/M70PbzPLBCGVZ9dollvj?p=preview

0. Set the browser to CST or any timezone less than GMT
1. Select 'Projects' in the left hand pane
2. Select 'Id' of any of the projects in the Project List; notice the 'Dev Date'
3. Now in the Edit Project form, the 'Development Date' displays decremented by 1 day
4. If user chooses to save, the 'Dev Date' will be stored in the db incorrectly.

I've tried it with and without the `.format` and `.label` with the same results. Again, creationView works just fine. And the datepicker requests with the `devdate:"2016-12-08"` so it is stored in the database accurately.

**Expected behavior:** Display 12-07-2016

**Actual behavior:** Displays 12-06-2016

matheo:
There's an infinite loop $watching value and rawValue, because the `Date` object adjust the `yyyy-mm-dd` value to the user's timezone, then modifying the original value +- the timezone. This patch adjust the timezone, so the rawValue is kept as the original value.

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maDateField.js
+++ b/src/javascripts/ng-admin/Crud/field/maDateField.js
@@ -29,7 +29,7 @@ export default function maDateField() {
                     return;
                 }
 
-                scope.rawValue = scope.value instanceof Date ? scope.value : new Date(scope.value);
+                scope.rawValue = scope.value instanceof Date ? scope.value : new Date(new Date(scope.value).getTime() + (new Date().getTimezoneOffset() * 60000));
             });
 
             scope.format = field.format();


Kmaschta:
There is an infinite loop, but it isn't due to the timezone but by a circular watch between two fields of the same scope.

I fix it with the PR #1318, which supersede this one.

Thanks for contributing!

matheo:
Infinite loop  caused by the timezone changing the original value, indeed, inside the circular watch. Once Date() changes the rawValue, it's all messed up, but the `angular.equals` check seems legit to prevent that too.

I confirmed that 1.0.1 works, thanks a lot!

It would be nice if we can get some help with `reference_many` #1311 and `embedded_list` #786 too please?
Best regards

PR #1318, opened 2017-03-08
[RFR] Fix DateField watches

### LINKED ISSUE #1271: editionView() returns date decremented by 1 day

### LINKED ISSUE #1305: Date field cannot be typed in

Kmaschta:
Related to #1271
Fix #1305
Supersede #1317

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ transpile:
 	cp -Rf ./src/sass/ lib/sass/
 
 build:
+	rm -rf build/*
 	make transpile
 	NODE_ENV=production ./node_modules/webpack/bin/webpack.js -p --optimize-minimize --optimize-occurence-order --optimize-dedupe --progress --devtool source-map
 	cp -Rf build examples/blog/
@@ -32,6 +33,9 @@ check-only-in-tests:
 
 test-unit:
 	./node_modules/.bin/karma start src/javascripts/test/karma.conf.js --single-run
+
+test-unit-watch:
+	./node_modules/.bin/karma start src/javascripts/test/karma.conf.js --watch
 
 test-e2e: prepare-test-e2e
 	./node_modules/.bin/protractor src/javascripts/test/protractor.conf.js
--- a/src/javascripts/ng-admin/Crud/field/maDateField.js
+++ b/src/javascripts/ng-admin/Crud/field/maDateField.js
@@ -1,3 +1,11 @@
+const valueToRawValue = (value) => {
+    if (value instanceof Date) {
+        return value;
+    }
+
+    return new Date(value);
+};
+
 /**
  * Edition field for a date - a text input with a datepicker.
  *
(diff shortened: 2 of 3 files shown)


matheo:
The persisting but minor issue is that the date selector also modify the original value I guess, because when you select a date it works fine, but if you open again, it show the previous date as selected. Maybe that plugin is having the timezone difference too.
This date-time stuff is quite sensible, Javascript itself needed more clarity handling these values.

Kmaschta:
I agree.
To be sure to sully understand you, can you open/complate an issue about that, with a simple example to reproduce it ?
Thanks

matheo:
I'm short of time right now,
and I guess that timezone issues will arise everywhere due to the multiple String <-> Date conversion on javascript. First on the DateField plugin, then later in the DatePicker. I'm experiencing issues with a Date filter too. So, this timezone stuff is all around.

thanhbuiea:
hi @Kmaschta , we still have problem with `Infinite Loop` due to `DateField watches`. I'm now on the latest version. This fix seems not to be able to resolve the bug. Could you please double check on it? If you need any clue or log, please let me know. Thanks. Hope to see you reply on this soon.

PR #1068, opened 2016-05-16
fixing issue #681

maksimbormot:
Fixes https://github.com/marmelab/ng-admin/issues/681

PR #1322, opened 2017-03-15
[RFR] Infinite scroll: avoid to load multiple time at the same time

### LINKED ISSUE #681: When not enough height, infinite pagination is never triggered
We must scroll a little bit to trigger the next page loading. Thus, when all the page is displayed without need for a scrollbar, we can't change page.

Kmaschta:
Fix #681
Supersedes #1068 

- [x] Use the native window `wheel` event
- [x] Only load the page if we down scroll (we loaded the page even with the top scroll)
- [x] Avoid to load a page if it's currently already loading it
- [x] Load at least one more page


package.json
```diff
@@ -52,6 +52,7 @@
     "babel-preset-stage-2": "^6.13.0",
     "chai": "^3.3.0",
     "css-loader": "^0.25.0",
+    "diff": "^3.2.0",
```
djhi: `^` are bad

Kmaschta: Even in dev deps ?

src/javascripts/ng-admin/Crud/list/maDatagridInfinitePagination.js
```diff
@@ -1,10 +1,14 @@
 import angular from 'angular';
 
-export default function maDatagridInfinitePagination($window, $document) {
+const isDownScrolling = wheelEvent => {
```
floo51: `isScrollingDown` ?

src/javascripts/ng-admin/Crud/list/maDatagridInfinitePagination.js
```diff
@@ -13,28 +17,59 @@ export default function maDatagridInfinitePagination($window, $document) {
...
+                            interval = null;
+                        }
                     }
-                    loadedPages.push(page);
-                    scope.nextPage()(page);
+                }, 100);
+            };
+
+            // Trigger the scroll at least once
+            // In this case, we'll avoid that a pagination of a few element (< 10)
```
floo51: Hard to understand, can you rephrase this or have a longer explanation ?

PR #1334, opened 2017-04-03
[RFR] Do not update date field actual value if invalid

### LINKED ISSUE #1305: Date field cannot be typed in

floo51:
Should fix #1305

### DIFF
--- a/src/javascripts/ng-admin/Crud/field/maDateField.js
+++ b/src/javascripts/ng-admin/Crud/field/maDateField.js
@@ -24,12 +24,14 @@ export default function maDateField() {
             scope.rawValue = scope.value == null ? null : valueToRawValue(scope.value);
 
             scope.$watch('rawValue', function(newRawValue) {
-                // ui bootstrap datepicker set value to undefined
-                // when the value is invalid and null if empty
-                const invalid = newRawValue === undefined;
+                if(newRawValue === undefined){
+                    // ui bootstrap datepicker set value to undefined
+                    // when the value is invalid and null if empty
+                    return;
+                }
                 const newValue = field.parse()(newRawValue);
 
-                if (!angular.equals(scope.value, newValue) && !invalid) {
+                if (!angular.equals(scope.value, newValue)) {
                     scope.value = newValue;
                 }
             });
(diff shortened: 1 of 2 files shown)


src/javascripts/ng-admin/Crud/field/maDateField.js
```diff
@@ -24,9 +24,12 @@ export default function maDateField() {
             scope.rawValue = scope.value == null ? null : valueToRawValue(scope.value);
 
             scope.$watch('rawValue', function(newRawValue) {
+                // ui bootstrap datepicker set value to undefined
+                // when the value is invalid and null if empty
+                const invalid = newRawValue === undefined;
```
Kmaschta: Can you `return` if invalid before the `field.parse`?

PR #1336, opened 2017-04-05
[RFR] Fix disabling views

floo51:
It should fix the disabling of views which wasn't working when you had fields initialized inside the view.

It also fix the test fixtures which were making tags a read only entity and trying to go to its edit page anyway.

### DIFF
--- a/examples/blog/config.js
+++ b/examples/blog/config.js
@@ -84,8 +84,7 @@
             .baseApiUrl('http://localhost:3000/') // The base API endpoint can be customized by entity
             .identifier(nga.field('id')); // you can optionally customize the identifier used in the api ('id' by default)
 
-        var tag = nga.entity('tags')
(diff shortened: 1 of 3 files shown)


floo51:
Mergeable - @Kmaschta

examples/blog/config.js
```diff
@@ -84,8 +84,7 @@
             .baseApiUrl('http://localhost:3000/') // The base API endpoint can be customized by entity
             .identifier(nga.field('id')); // you can optionally customize the identifier used in the api ('id' by default)
 
-        var tag = nga.entity('tags')
-            .readOnly(); // a readOnly entity has disabled creation, edition, and deletion views
```
Kmaschta: What are the effects of your `admin-config` update on `readOnly` feature?

floo51: The effect is that it now works

PR #1339, opened 2017-04-25
[RFR] Use http error service handlers inside view events

floo51:
The `HttpErrorService` decorator wasn't working on view like on "Submit changes".

It should now work.

Still untested, need some more work.


Kmaschta:
I'm OK with it.
@Phocea Do you have any observation?

floo51:
Just tested, works as expected

PR #1346, opened 2017-06-13
Fix Child Menu Highlight

Phocea:
With previous code the parent menu gets highlighted only when the first child is active and not when any of its children are active.
This change fixes that.


Phocea:
Actually I am revising my idea on this.
Currently active is used to checked on the parent, if the current url include the given path. In my use case, all of the child menu item have a different path with no common denominator.
So using menu.isActive(path) would only work on a given one... fix for me was to use childmenu.isActive(path), but I suppose this could be done in a template, depending on the need ?

Kaban-1:
[128a094153d64ff5e230-cf0203d69fc74c95d093885b4602ed69c474e2c4.zip](https://github.com/marmelab/ng-admin/files/1275933/128a094153d64ff5e230-cf0203d69fc74c95d093885b4602ed69c474e2c4.zip)

Issue #1106, opened 2016-06-07
deletionView and onSubmitSuccess/onSubmitError


tomy2105:
I think deletion view should have those two methods too.
This would allow you to have some custom actions after delete is done on server or custom handling if delete fails on server.

PR #1111, opened 2016-06-08
onSubmit functions and progress added to DeleteView along with embedded list button fix

tomy2105:
DeleteController changes so onSubmit functions added to DeleteView are invoked.
Progress stat/stop invoked.

admin-config pull request https://github.com/marmelab/admin-config/pull/67 needed for this to function!

This request is related to:
https://github.com/marmelab/ng-admin/issues/1106
and
https://github.com/marmelab/ng-admin/issues/1109
and
https://github.com/marmelab/ng-admin/issues/1114

PR #1361, opened 2017-09-05
deletionView and onSubmitSuccess/onSubmitError

tomy2105:
This resolves https://github.com/marmelab/ng-admin/issues/1106 and obsoletes (already closed) https://github.com/marmelab/ng-admin/pull/1111.


tomy2105:
Due to unfortunate usage of bash on windows at the moment (need some time to build my linux machine) I cannot run tests and thus unable to fix test case to follow my changes.
I've tested manually and my fix works, so I'm pretty sure it is only test case that needs update.
If anyone else might give me a hand with this I'd be grateful.

tomy2105:
OK, the initial error due to test case not being update is gone. But again I don't understand check failed error message "No selenium server jar found at", can someone please see what happened and why please.

Issue #1204, opened 2016-09-20
Filter field wont show defaultValue when pinned


avifatal:
this will not set default 

```
var isa = nga.field('is_archived','boolean');
isa.defaultValue(false);
isa.pinned(true);

....filters([ isa ] )
```

Phocea:
Looks like this is re-occuring as demonstrated in this plunker: http://plnkr.co/edit/9qJ0wcSOzCFFDLe31U1K?p=preview

setting a filter with defaultValue(true) on  the User listview is ok, but defaultValue(false) does not select the false boolean value...

PR #1366, opened 2017-09-29
[RFR] Update filters on view initialization

floo51:
Fixes an issue where a pinned filter wouldn't filter with its initially selected value.


Kmaschta:
Does this PR fix this issue ? https://github.com/marmelab/ng-admin/issues/1204

floo51:
No, I don't think `defaultValue` actually applies for filters.
We can still implement something like it with `map` and `transform` returning our default value where no `value` is given in the map function by instance.

PR #1394, opened 2019-01-03
Updated many libs. Fixed embedded list controls validation. Changed s…

StefanNedelchev:
…ubmit button in edit pages to be disable when the form is $pristine.


StefanNedelchev:
updates

